PR #2, opened 2015-04-27
Update of FFV1 specification, Header part

JeromeMartinez:
This is the first patch of a serie of patches having the goal of being able to create a decoder only based on this specification, without having to rely on the FFV1 encoder/decoder implemented in FFmpeg. This document aims to be the base document for our proposal of FFV1 standardisation.
We currently focused on the header part of this document.

### DIFF
--- a/ffv1.lyx
+++ b/ffv1.lyx
@@ -4160,33 +4160,390 @@ u(32)
 
 \begin_layout Labeling
 \labelwidthstring 12345678901234567890
-slice_coding_mode 0: normal RC or VLC
+slice_coding_mode indicates the slice coding mode.
 \begin_inset Newline newline
 \end_inset
 
-1: store slice samples as raw PCM instead of context adaptive range coding
+
+\begin_inset Tabular
+<lyxtabular version="3" rows="4" columns="2">
+<features rotate="0" tabularvalignment="middle">
+<column alignment="center" valignment="top">
+<column alignment="center" valignment="top">
+<row>
+<cell alignment="center" valignment="top" topline="true" bottomline="true" leftline="true" usebox="none">
+\begin_inset Text
+
+\begin_layout Plain Layout
+value
+\end_layout
+
+\end_inset
+</cell>
+<cell alignment="center" valignment="top" topline="true" bottomline="true" leftline="true" rightline="true" usebox="none">
+\begin_inset Text
+
+\begin_layout Plain Layout
+slice coding mode
+\end_layout
+
+\end_inset
+</cell>
+</row>
+<row>
+<cell alignment="center" valignment="top" topline="true" leftline="true" usebox="none">
+\begin_inset Text
+
+\begin_layout Plain Layout
+0
+\end_layout
+
+\end_inset
+</cell>
+<cell alignment="center" valignment="top" topline="true" leftline="true" rightline="true" usebox="none">
+\begin_inset Text
+
+\begin_layout Plain Layout
+normal Range Coding or VLC
+\end_layout
+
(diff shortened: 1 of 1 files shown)


michaelni:
merged the first 2 comits, please be more careful about the commit messages
"sg Golomb Rice coded unsigned" sg unsigned ?
also maybe whatever-rle should be used instead of sl, the term rle is widely known
also please watch out for trailing whitespace, many lines of the commits  add trailing whitespace

michaelni:
RFC 2119 uses all uppercase terms, this should not be changed
other parts of the spec use lists or whatever, the RFC terms would be added as a table
reference go to the references section

michaelni:
the remainder of the patches dont seem to apply cleanly without the previous ones
also i somehow think review per email would be easier than using github

TimothyGu:
@michaelni you can add a line note on GitHub by hovering your mouse over the lines in "files changed" tab and press the blue + on the left.

JeromeMartinez:
> the remainder of the patches dont seem to apply cleanly without the previous ones

Right, I moved some parts of the specs and it is not possible without previous patches.

> also i somehow think review per email would be easier than using github 

per email, do you mean direct to your email address or ffmpeg-devel list?

JeromeMartinez:
> RFC 2119 uses all uppercase terms, this should not be changed

the initial spec was using lower cases and I did not want to change that, I'll update with all in uppercase.

JeromeMartinez:
I corrected the issue signed/unsigned
I put rle for Run Length Encoding
No additional unnecessary trailing whitespace (the ones which are present are added by Lyx and are actually necessary for the formating)
RFC 2119 with uppercase and with a link added to the reference section.

Please review the new patches.
I removed the other ones in order to have a clean first PR and I'll create a new PR (or email to ffmpeg-devel if you prefer)

TimothyGu:
The patches still don't apply cleanly. Please rebase to current master.

JeromeMartinez:
Please accept my apologies about the lack of rebase, it is done now and patches should apply cleanly now ("This pull request can be automatically merged" from GitHub)

michaelni:
lre typo in commit message, and the added "rle" seems not used anywhere, also maybe it should be srle for consistency with s/u signed/unsigned prefixes

JeromeMartinez:
You're right, rle is not yet used, I added it by anticipation.
For the moment, I remove it from my patch and I'll re-introduce it when I really need it (if I need it, I'll see how I describe golomb rice).

Please review again.

michaelni:
applied the accronyms patch, about the reformating one i think that is the wrong structure. Before it used "Labeling" which seperated the identifer from the text describng the identifer. After the patch it would be a "itemize" where the seperation is merly through a newline, this doesnt feel semantically correct

JeromeMartinez:
I switched back to "Labeling".
Unfortunately, the HTML output from Lyx is difficult to read when there are complex "Labeling" blocks (e.g. with carriage return or arrays). I propose a second patch for switching to "Description" which looks like semantically correct and is better displayed in HTML (which will be the one people look at).

Please review again

michaelni:
the text for micro version should make it more clear that micro versions depend on the version so that for version 5 micro version could start again at 0

JeromeMartinez:
in this patch, I wanted to reformat the current content, without changing the meaning or precision of the previous text (split of patches in several parts). Another patch is in the pipe with more precision about micro_versio, is possible to commit the current patch as is (only some reformating, no content changing) and I send another patch with more details about micro_version?

michaelni:
all commits applied

PR #4, opened 2015-04-28
More details about some fields

JeromeMartinez:
More details about version, micro_version, bits_per_raw_sample, crc_parity fields and Quantization Tables are provided, in order ot have an explicit definition of each field.

### DIFF
--- a/ffv1.lyx
+++ b/ffv1.lyx
@@ -5246,7 +5246,7 @@ br
 \begin_inset space ~
 \end_inset
 
-QuantizationTables
+QuantizationTable(0)
 \end_layout
 
 \end_inset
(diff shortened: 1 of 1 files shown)


michaelni:
first 3 commits merged, the quanti table comit contains inconsistet formating, a "int a" but no types elsewhere, sometimes spaces around operators sometimes not, i think it should be made consistent
also you asked about postig to ffmpeg-devel, maybe this would be a good idea, others possibly would spot formating inconsistencies quicker and more completely than i would

JeromeMartinez:
Formating (including space) is now consistent with other parts of the specs, type removed.

I can also post it on ffmpeg-devel.

michaelni:
i suggest, yes lets try to post it to ffmpeg-devel and see how that works we can return to using github if its worse. Either please always also push to your github branch, it makes merging slightly easier than applying the patch from the ML

PR #10, opened 2015-07-10
transition to markdown, readme, and derivatives

dericed:

PR #13, opened 2015-07-21
move code from tables to codeblocks

dericed:
As mentioned in https://github.com/FFmpeg/FFV1/pull/10#issuecomment-122553817, this PR moves code from tables to commented code blocks.

### DIFF
--- a/ffv1.md
+++ b/ffv1.md
@@ -436,7 +436,7 @@ Note, this is different from JPEG-LS, which doesn’t use prediction in run mode
 
 # Bitstream
 
-|      |                                                        |
+|Symbol| Defintion                                              |
 |------|--------------------------------------------------------|
 | u(n) | unsigned big endian integer using n bits               |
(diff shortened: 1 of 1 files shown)


michaelni:
the output of this is full of backslashes in pdf
btw, the table style originates from ISO mpeg specs

ablwr:
This PR is coming up on a year old. Resolve or remove?

dericed:
I'm going to close. I think resolving the conflict may be more trouble that it's worth.

ablwr:
If still an issue, let's open up an issue!

dericed:
I think it the content of where the document is now, the approach to the issue is different. For instance in https://datatracker.ietf.org/doc/draft-niedermayer-cellar-ffv1/?include_text=1 see sections 5.5, 5.6, 5.7 as the tables don't render. I think the code looks okay in tables, but most of them should be adjusted to be left-aligned.

PR #16, opened 2015-12-02
Typos and cleanup

dericed:
minor changes: typos, grammar, formatting

### DIFF
--- a/ffv1.md
+++ b/ffv1.md
@@ -218,11 +218,11 @@ $coder\_input=\left[\left(sample\_difference+2^{bits-1}\right)\&\left(2^{bits}-1
 
 ### Range coding mode
 
-Early experimental versions of FFV1 used the CABAC Arithmetic coder from [H.264](#references) but due to the uncertain patent/royality situation, as well as its slightly worse performance, CABAC was replaced by a range coder based on an algorithm defined by *G. Nigel N. Martin* in 1979 [RangeCoder](#references).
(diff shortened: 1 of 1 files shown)


michaelni:
"Each frame is composed of 1 to 4 of the following colorspace planes: Y, Cb, Cr, Alpha"
This is true for YCbCr only

dericed:
Right. I also think this section needs to be changed to reflect the hierarchy of frame>slice>plane, as currently this paragraph only refers to frame and plane, and then slice arrives later in the documentation without a good introduction.

dericed:
I'm reviewing this again and have some questions:

- When the colorspace is RCT, are planes used? When I encode rgb24 with ffv1.v3 I only see one plane, but may be misinterpreting this.
- Re: "Each frame is composed of 1 to 4 ... planes". I'd like to be more specific. From this reading it seems like 2 planes is possible but that Cb should follow Y; however, are `ya` colorspaces possible (grayscale with alpha)?
- Is a 2 plane encoding possible?
- What is the meaning of `line` in this paragraph? The line count appears effected by the colorspace and possibly the chroma subsampling, but I'm not clear about this.
- Re: "since it is most likely that the RCT will immediately be converted to RGB during decoding" what other conversion is possible?

michaelni:
i dont think pull requests are the right place for such questions, theres a IETF and a ffmpeg mailing list.
Here in the pull req basically only 2 people will see it

dericed:
I moved the questions to an email here: https://mailarchive.ietf.org/arch/msg/cellar/nw5nwEmfnEEBu88QSOyJbsOQu5E. It appears the commits of this PR are already in the repository so I am closing the PR.

PR #25, opened 2016-12-04
Use rfc formatting

dericed:
This pull request updates many aspects of the formatting for better handling when transformed through rfc2xml. For instance see the visual issues with https://tools.ietf.org/html/draft-niedermayer-cellar-ffv1-00#section-5.2.

This PR is not intended to produce any semantic change, except for some minor changes in referencing.

@michaelni please confirm that the reference to jpeg2000.org can be
swapped to ISO.15444-1.2016. Preferable to reference a standard rather
than a website.

The jpegls ref is updated to iso14495. The fcd14495p.pdf url was dead anyhow.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,11 +8,11 @@ $(info RFC rendering has been tested with mmark version 1.3.4 and xml2rfc 2.5.1,
 all: ffv1.html ffv1.pdf draft-niedermayer-cellar-ffv1-00.html draft-niedermayer-cellar-ffv1-00.txt
 	
 ffv1.html: ffv1.md
-	cat pdf_frontmatter.md "$<" > merged.md
+	cat pdf_frontmatter.md "$<" pdf_backmatter.md > merged.md
 	pandoc --toc --mathml -s --number-sections  -c "style.css" -o "$@" merged.md
 
 ffv1.pdf:  ffv1.md
-	cat pdf_frontmatter.md "$<" > merged.md
+	cat pdf_frontmatter.md "$<" pdf_backmatter.md | sed "s|\[@!|\[|g;s|\[@?|\[|g;s|\[@|\[|g" > merged.md
 	pandoc --toc -s --number-sections --latex-engine=xelatex -V geometry:margin=1in -o "$@" merged.md
 
 draft-niedermayer-cellar-ffv1-00.html: ffv1.md
--- a/ffv1.md
+++ b/ffv1.md
@@ -8,7 +8,7 @@ This document assumes familiarity with mathematical and coding concepts such as 
 
 # Notation and Conventions
 
-The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [RFC 2119](https://tools.ietf.org/html/rfc2119).
+The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "SHOULD NOT", "RECOMMENDED", "MAY", and "OPTIONAL" in this document are to be interpreted as described in [@!RFC2119].
 
 ## Definitions
 
@@ -38,7 +38,7 @@ TBA    To Be Announced. Used in reference to the development of future iteration
 
 ## Conventions
 
-Note: the operators and the order of precedence are the same as used in the C programming language [ISO/IEC 9899](#references).
+Note: the operators and the order of precedence are the same as used in the C programming language [@!ISO.9899.1990].
 
 ### Arithmetic operators
 
@@ -165,7 +165,7 @@ median(left, top, left + top - diag)
 
 left, top, diag are the left, top and left-top samples
 
-Note, this is also used in [JPEG-LS and HuffYuv](#references).
+Note, this is also used in [@ISO.14495-1.1999] and [@HuffYUV].
 
 Exception for the media predictor:
 if colorspace_type == 0 && bits_per_raw_sample == 16 && ( coder_type == 1 || coder_type == 2 ), the following media predictor MUST be used:
@@ -215,7 +215,7 @@ $r=Cr+g$
 
 $b=Cb+g$
 
-[JPEG2000](#references)
+[@!ISO.15444-1.2016]
 
 ## Coding of the sample difference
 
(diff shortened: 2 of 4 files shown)


ablwr:
Very great! 👍

retokromer:
The standard is officially called `ISO/IEC 15444-1:2016`. I would not drop the IEC part.

In the section _Assignment operators_ I would write `a := a + 1` etc. instead of `a = a + 1` etc.

dericed:
@retokromer I'm following the practice for referencing ISO/IEC standards as done at https://xml2rfc.tools.ietf.org/public/rfc/bibxml2/. In the cases here ISO/IEC standards use anchors that are prefixed with only `ISO.`.

michaelni:
The change for QuantizationTablePerContext adds some \
which show up in the final output, these where not there before, are they intended?

michaelni:
the indention of the elements in the 2nd column is inconsistet, for example the br element between the Frame and SliceHeader

michaelni:
in SliceHeader() you randomize the indention, sar_den and the following if() are at the same indentio level before a02463a8147b353fc023ef22107dfb44a7c824ed and 1 space different after.

dericed:
@michaelni I couldn't see the character you meant in https://github.com/FFmpeg/FFV1/pull/25#issuecomment-264713054. But I found an extra `|` character in there and removed it in https://github.com/FFmpeg/FFV1/pull/25/commits/fed0e6fea7a68a8ef6de5eb7e69b10819a614cb6.

dericed:
Indentation noted in https://github.com/FFmpeg/FFV1/pull/25#issuecomment-264713264 is fixed in https://github.com/FFmpeg/FFV1/pull/25/commits/16246362f4826a2f806efa4a37bd27612e41bf1f.

dericed:
I think I've updated according to all comments. Please re-review. I can also squash the codeblock commits together if easier.

michaelni:
Comment 25 was about a backslash
about combining and spliting commits, please combine bugfixes with commits introducing the bug so no known issues are added while keeping commits seperated. Note this would be much easier if you submit things early as then little rebasing of later commits would be needed

dericed:
Commits are squashed to eliminate introducing/fixing bugs and for readability.

michaelni:
Thanks
indention in Parameters() is random, 
";" placement at end of lines is random too everywehre 
space placement in for() / while() and if() mismatches
space placement in function args and () is inconsistent
space placement around array indexes is inconsistent
Most of this exists priot to the patch some is made worse by it though like the indention in Parameters()

retokromer:
@dericed It´s not the link itself, but the wording I see that hurts me: the title reads wrong in markdown this way.

dericed:
I updated the Parameters indentation. For the rest which exists prior, let's leave for a later PR.

dericed:
@retokromer can you move the question on iso citation style to cellar.

michaelni:
The change to SliceContent() is not correct, same for Line(), the indention in Parameters() also is still not correct

dericed:
I reviewed codeblocks for indentation and order. PR is updated.

michaelni:
Pixel() is wrongly indented after te first patch in Line()

dericed:
I fixed `Pixel()` indentation in `Line()` and resolved merge conflicts.

dericed:
And rebased to remove the commit on resolving merge conflicts. Should be ready.

michaelni:
merged 1 patch, Is it intended that the "left-align" patch skips the Suffix Examples table ?

dericed:
Updated to left-align the Suffix Examples table.

dericed:
@retokromer, regarding https://github.com/FFmpeg/FFV1/pull/25#issuecomment-264687813, I agree but didn't want to extend the scope of the PR, so I filed it under https://github.com/FFmpeg/FFV1/issues/27

dericed:
I added a [commit](https://github.com/FFmpeg/FFV1/pull/25/commits/e1becfbe537759a068d45304df1509acefb844b1) to get the references back to the `ffv1.pdf` since that output doesn't gather them automatically from the xml2rfc process. If of interest, then I can merge that with https://github.com/FFmpeg/FFV1/pull/25/commits/f6d5db0a9c8afe1a5501fc2f1c3b67e920d3036e.

dericed:
@michaelni, anything else needed for the remaining commits?

michaelni:
4504e25e9bb2ad79840f0455672fe58f1c26536d make the table rather hard to read in the pdf. The individual terms are hard to see in the flowing text that reaches into the same first column

dericed:
https://github.com/FFmpeg/FFV1/commit/4504e25e9bb2ad79840f0455672fe58f1c26536d is updated to https://github.com/FFmpeg/FFV1/pull/25/commits/a7cedb077efc1e4103bf1cfe96515be1441ae943, which is a better compromise between the output of mmark/xml2rfc and pandoc.

michaelni:
Is it intended that a130bbceba6365c57d7d63115a78c7873eefa1e6 causes litteral "@!" to be shown in the the output (pdf in this case) ?

dericed:
The "@!" is part of mmark's syntax to identify a reference in the RFC output, though this shows as a literal in the PDF output. I suppose we could add a 'sed s/[@!/[/g' to the Makefile before the PDF output is rendered. ?

michaelni:
On Wed, Dec 14, 2016 at 05:34:26AM -0800, Dave Rice wrote:
> The "@!" is part of mmark's syntax to identify a reference in the RFC output, though this shows as a literal in the PDF output. I suppose we could add a 'sed s/[@!/[/g' to the Makefile before the PDF output is rendered. ?

if theres no cleaner solution then that seems like a good idea

thx

[...]
-- 
Michael     GnuPG fingerprint: 9FF2128B147EF6730BADF133611EC787040B0FAB

I have often repented speaking, but never of holding my tongue.
-- Xenocrates

retokromer:
+1 for filtering out mmark’s own syntax before rendering the PDF.

dericed:
The solution suggested in https://github.com/FFmpeg/FFV1/pull/25#issuecomment-267034792 is implemented in https://github.com/FFmpeg/FFV1/pull/25/commits/ccd33f284b912c5f8381d129a7ac5feb71fc21ed#diff-b67911656ef5d18c4ae36cb6741b7965R15. I also dropped a few commits from the PR that were already cherrypicked.

retokromer:
LGTM

michaelni:
merged

PR #28, opened 2016-12-14
add subsections

dericed:
This PR add subsections for defined terms, particularly for components on line, frame, slice, and parameters. By having section headers for these terms we can use forward referencing to them. For instance in the slice section, `coder_type` is used before it is defined so it could use a forward reference. Comments welcome.


michaelni:
commit merged

PR #29, opened 2016-12-14
WIP narrative explanation on Slice

dericed:
I think it may make the document more readable if some of the structure which is presently defined in code blocks is accompanied by a narrative description. Here’s an example of that approach for slice. This also includes an example of forward-referencing regarding ‘version’. Note: these references appear differently between the pandoc and mmark/rfc2xml outputs.
Comments welcome.

### DIFF
--- a/ffv1.md
+++ b/ffv1.md
@@ -537,6 +537,13 @@ Frame( ) {                                                    |
 
 ## Slice
 
+A `Slice` is a spatially distinct region of a `Frame` that is encoded separately from any other region of that `Frame`. A `Slice` is comprised the following four components in this storage order:
+
+- `SliceHeader` (Present if the `version` (see [the section on version](#version)) is greater than or equal to `3`)
+- `SliceContent` (MUST be present)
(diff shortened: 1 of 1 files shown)


dericed:
Thoughts on this approach. I think adding such narratives could be helpful to some readers, though it adds a risk of conflict between the narrative and the code. As an alternate we could better define how to read the code. I don't want to presume too much that it is intuitive.

dericed:
Going to drop this PR in favor of offering some context in the document to explain that C-like code-blocks are used to explain the storage of the components of FFV1.

retokromer:
+1
Yet why «C-like»? Is it not possible to use plain C code-fragments?

JeromeMartinez:
> Yet why «C-like»? Is it not possible to use plain C code-fragments?

plain C is more verbose and more difficult to read, and permits lot of traps.
Additionally, plain C adds more constraints (e.g. is the integer type 8-bit or 16-bit? how do you handle integer overflows? it should depend of your design choice,) the spec should be neutral (e.g. consider number are unlimited)

retokromer:
I understand your arguments, but I don’t agree with you, except the increased verbosity.

BTW: In my teaching experience, Pascal-style pseudo code is often considered more intuitive than C-style pseudo code by the students.

PR #32, opened 2016-12-18
Standardize some terms

dericed:


dericed:
ping

michaelni:
Just merged but please use your full name in the author field not dericed. This should make it easier for people to know who wrote what, unless you dont want your name in there

Issue #34, opened 2017-01-03
define NumBytes


dericed:
Currently `NumBytes` is defined in an aside like `The size of the Configuration Record, NumBytes, is supplied by the underlying container.` but `NumBytes` via `remaining_bits_in_bitstream` is also needed in context with `Frame`. I suggest that `NumBytes` have it's own definition, as I was initially confused to think that `NumBytes` was only a measurement of the Configuration Record and not of a Frame.

dericed:
Closed by https://github.com/FFmpeg/FFV1/commit/2e864f6f4a537636a95a3e59114874a144168bae

PR #37, opened 2017-01-07
add NumBytes definition

dericed:
Also adds a header for container mappings so that I could call that in a forward-reference.


dericed:
intended to resolve https://github.com/FFmpeg/FFV1/issues/34

Issue #14, opened 2015-12-01
add citations to prerequisite documents


dericed:
The https://github.com/FFmpeg/FFV1/blob/master/ffv1.md#introduction section refers to familiarity with concepts of Range coding and YCBCr colorspaces. I propose that we link to or name specific resources that can provide this familiarity.

michaelni:
sure, do you have a patch or suggestions for which resources to link to ?

dericed:
Not yet, for now I'm just trying to log issues for things to do. I'll look for a good resources on these issues that isn't behind a paywall.

PR #39, opened 2017-01-07
add references to introduction

dericed:
intended to resolve https://github.com/FFmpeg/FFV1/issues/14

### DIFF
--- a/ffv1.md
+++ b/ffv1.md
@@ -4,7 +4,7 @@ The FFV1 video codec is a simple and efficient lossless intra-frame only codec.
 
 The latest version of this document is available at <https://raw.github.com/FFmpeg/FFV1/master/ffv1.md>
 
-This document assumes familiarity with mathematical and coding concepts such as Range coding and YCbCr colorspaces.
+This document assumes familiarity with mathematical and coding concepts such as Range coding [@?range-coding] and YCbCr colorspaces [@?YCbCr].
 
 # Notation and Conventions
 
--- a/pdf_backmatter.md
+++ b/pdf_backmatter.md
@@ -34,3 +34,5 @@ ASAN Addresss Sanitizer, <http://clang.llvm.org/docs/AddressSanitizer.html>.
 REFIMPL, The reference FFV1 implementation / the FFV1 codec in FFmpeg, <https://ffmpeg.org/>.
 
 OPUS, Definition of the Opus Audio Codec, <https://www.ietf.org/rfc/rfc6716.txt>
+
+YCbCr, Wikipedia, "YCbCr", <https://en.wikipedia.org/w/index.php?title=YCbCr>.
--- a/rfc_frontmatter.md
+++ b/rfc_frontmatter.md
@@ -120,4 +120,14 @@ This document defines FFV1, a lossless intra-frame video encoding format. FFV1 i
   </front>
 </reference>
 
+<reference anchor="YCbCr" target="https://en.wikipedia.org/w/index.php?title=YCbCr">
+  <front>
+    <title>YCbCr</title>
+    <author>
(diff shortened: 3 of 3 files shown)


michaelni:
Bt601 does not appear in the reference section in the pdf

retokromer:
Do you mean [Rec. 709](https://www.itu.int/dms_pubrec/itu-r/rec/bt/R-REC-BT.709-6-201506-I!!PDF-E.pdf)?

JeromeMartinez:
pointing on 1 specific document which provides choices about transformations constants is maybe not the best choice (why BT.601 and not BT.709 or BT.2020?)
Maybe a more generic paper on color spaces?

retokromer:
It seems to me that Rec. 709 fits better the purpose than Rec. 601, but I plenty agree with @JeromeMartinez. I would recommend Charles Poynton (I also use Y′C<sub>B</sub>C<sub>R</sub> myself).

dericed:
- Updated from bt601 to bt709.
- Added the same reference to the pdf references section.

Ready for review again.

michaelni:
Refering to bt709 would suggest that no other YCbCr can be stored or that its implied to be bt709 "compatible"

JeromeMartinez:
> Refering to bt709 would suggest that no other YCbCr can be stored or that its implied to be bt709 "compatible"

None of them, as bt709 is only one of the possibilities for YCbCr.
@dericed, why a specific standard (here, HDTV, excluding SDTV or UHD or...)? From my point of view, it may be misleading, as the bt number is in the Matroska for the moment (we plan to add the bt number later in FFV1 as a metadata)

dericed:
Since the intro suggests that familiarity with YCbCr is assumed, I wanted to add a reference in case the reader needs that info.

JeromeMartinez:
Having a reference is good, but having a specific YCbCr transformation as a reference may make people think that this is the "standard" method for FFV1 (and it is not).
As the constants are not provided by FFV1 (external metadata for the moment), I suggest something more neutral e.g. https://en.wikipedia.org/wiki/YCbCr which says that there are several possibilities.

retokromer:
I suggested Rec. 709 because in my experience is what people request most those days. The Wikipedia entry is not completely correct, but Charles Poynton’s attemps to improve it failed. I’m OK with it, not less not more.

JeromeMartinez:
> I suggested Rec. 709 because in my experience is what people request most those days.

Really depends of who you are talking to. In some other domains, this is [XYZ](https://en.wikipedia.org/wiki/CIE_1931_color_space) (cinema, and they also use lossless video, but ok maybe out of topic it is not YCbCr) or [HLG](https://en.wikipedia.org/wiki/Hybrid_Log-Gamma) (broadcast, extending YCbCr from BT.2020 used for 4KK, and there are more and more 4K content).
My concern is to not be focused on what we use to have now in our domain of expertise and to try to stay neutral. Yesterday was BT.601, today is BT.709, tomorrow is BT.2020/HLG, the day after tomorrow will be something else. And yourself use other color spaces. Wikipedia entry is just an idea, my concern is to avoid to focus on a particular color space, as it may be misleading as an endorsement of this particular color space in FFV1.

If we stay with a particular BT.xxx, maybe writing that this is only an example and that there is no endorsement of particular YCbCr transfer constants.

retokromer:
> And yourself use other color spaces.

Indeed! And I know why… ;-)

The colour space XYZ is the origin which is meant to represent scientifically the view of an average trichromat.

As said, I’m OK with the Wikipedia entry for Y′C<sub>B</sub>C<sub>R</sub>.

dericed:
Updated from ITU bt709 to Wikipedia:YCbCr

retokromer:
LGTM

PR #49, opened 2017-02-11
add Matroska mapping

dericed:


dericed:
should match with https://github.com/Matroska-Org/matroska-specification/blob/master/codec_specs.md#v_ffv1

JeromeMartinez:
I think that the "V_MS/VFW/FOURCC" part should not be there: content of the paragraph about the ConfigurationRecord position is already in the AVI part, and "V_MS/VFW/FOURCC" part in Matroska is for AVI compatibility.
Here, I see a mix of Matroska and AVI details in the same paragraph, not good.

I suggest that the 2nd paragprah of FFV1 in Matroska is removed, alternatively with the BITMAPINFOHEADER part as another PR about extending AVI part.

dericed:
updated per @JeromeMartinez's comments, though I want to be sure that someone using this spec to interpret ffv1 in mkv, doesn't ignore all the legacy mkv with ffv1 in VFW

michaelni:
Maybe you want to add some text mentioning VFW

PR #50, opened 2017-02-12
add method to use latex(pdf) or ascii(rfc) for math

dericed:
prefix a line with either “PDF:” or “RFC:”, the prefixes will be
removed and the lines that don’t match the output will be dropped. This
allowed support of both latex and ascii outputs. Comments? Particularly
about the use of the seq() function?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,20 +8,20 @@ $(info RFC rendering has been tested with mmark version 1.3.4 and xml2rfc 2.5.1,
 all: ffv1.html ffv1.pdf draft-niedermayer-cellar-ffv1-01.html draft-niedermayer-cellar-ffv1-01.txt
 	
 ffv1.html: ffv1.md
-	cat pdf_frontmatter.md "$<" pdf_backmatter.md > merged_ffv1html.md
+	cat pdf_frontmatter.md "$<" pdf_backmatter.md | grep -v "^RFC:" | sed "s|^PDF:||g" > merged_ffv1html.md
 	pandoc --toc --mathml -s --number-sections  -c "style.css" -o "$@" merged_ffv1html.md
 
 ffv1.pdf:  ffv1.md
-	cat pdf_frontmatter.md "$<" pdf_backmatter.md | sed "s|\[@!|\[|g;s|\[@?|\[|g;s|\[@|\[|g" > merged_ffv1pdf.md
+	cat pdf_frontmatter.md "$<" pdf_backmatter.md | grep -v "^RFC:" | sed "s|\[@!|\[|g;s|\[@?|\[|g;s|\[@|\[|g;s|^PDF:||g" > merged_ffv1pdf.md
 	pandoc --toc -s --number-sections --latex-engine=xelatex -V geometry:margin=1in --variable urlcolor=blue -o "$@" merged_ffv1pdf.md
 
 draft-niedermayer-cellar-ffv1-01.html: ffv1.md
-	cat rfc_frontmatter.md "$<" > merged_rfchtml.md
+	cat rfc_frontmatter.md "$<" | grep -v "^PDF:" | sed "s|^RFC:||g" > merged_rfchtml.md
 	mmark -xml2 -page merged_rfchtml.md > draft-niedermayer-cellar-ffv1-01.xml
 	xml2rfc --html draft-niedermayer-cellar-ffv1-01.xml -o "$@"
 
 draft-niedermayer-cellar-ffv1-01.txt: ffv1.md
-	cat rfc_frontmatter.md "$<" > merged_rfctxt.md
+	cat rfc_frontmatter.md "$<" | grep -v "^PDF:" | sed "s|^RFC:||g" > merged_rfctxt.md
 	mmark -xml2 -page merged_rfctxt.md > draft-niedermayer-cellar-ffv1-01.xml
 	xml2rfc draft-niedermayer-cellar-ffv1-01.xml -o "$@"
 
--- a/ffv1.md
+++ b/ffv1.md
@@ -97,15 +97,21 @@ Note: the operators and the order of precedence are the same as used in the C pr
 ### Mathematical functions
 
 --------------------- -----------------------------------------------
-$$\lfloor a \rfloor$$ the largest integer less than or equal to a
+PDF:$$\lfloor a \rfloor$$ the largest integer less than or equal to a
+RFC:floor(a)              the largest integer less than or equal to a
 
-$$\lceil a \rceil$$   the smallest integer greater than or equal to a
+PDF:$$\lceil a \rceil$$   the smallest integer greater than or equal to a
+RFC:ceil(a)               the largest integer less than or equal to a
 
 abs(a)                the absolute value of a, i.e. abs(a) = sign(a)*a
 
 log2(a)               the base-two logarithm of a
 
 min(a,b)              the smallest of two values a and b
+
+RFC:a_{b}                 the b-th value of a sequence of a
+RFC:
+RFC:a_{b,c}               the 'b,c'-th value of a sequence of a
 --------------------- -----------------------------------------------
 
 ### Order of operation precedence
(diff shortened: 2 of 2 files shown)


Makefile
```diff
@@ -8,20 +8,20 @@ $(info RFC rendering has been tested with mmark version 1.3.4 and xml2rfc 2.5.1,
 all: ffv1.html ffv1.pdf draft-niedermayer-cellar-ffv1-01.html draft-niedermayer-cellar-ffv1-01.txt
 	
 ffv1.html: ffv1.md
-	cat pdf_frontmatter.md "$<" pdf_backmatter.md > merged_ffv1html.md
+	cat pdf_frontmatter.md "$<" pdf_backmatter.md | grep -v "^RFC:" | sed "s|PDF:||g" > merged_ffv1html.md
```
TimothyGu: `^PDF:` (and also below)

dericed:
I updated the grep/sed process per @TimothyGu's comment. I also added code-block ticks around the equations in the RFC presentation (I think they look better in the txt and html rfc output this way).

dericed:
Some questions based on this.

The "Range binary values" section contains the most complex math:

- in the RFC version is it more clear to have a "^" as a mathematical wedge or is it preferable to use "AND"?

- are "<==" and "<==>" acceptable ascii versions of \Longleftarrow and \Longleftrightarrow?

In the Context section:

- this part includes several absolute values in `|Q|` format. When this is expressed in ascii rather than LaTeX with pipes, is the meaning still clear or would it be better to replace these with an `abs()` function?

michaelni:
"^" to me is pow() or XOR, The arrows and |Q| seem fine to me, i dont know how others see all these ...

dericed:
I reviewed this more closely and force pushed an update. I caught a few typos and replace the "^" with XOR. I also found two instances in the latex equations were slice_x and slice_y were producing the y and x in subscript. I fixed these by using `slice\_x` (escaping the underscore) as needed in latex in markdown.

I removed the "WIP" from the title of the pull request as I'm reasonable certain that this is okay. The one area I'm hesitant about is using a new `seq()` function to replace the use of subscript in the latex equations. It renders the "Range binary values" section like this:

```
3.6.1.1.  Range binary values

   To encode binary digits efficiently a Range coder is used. "seq(C,i)"
   is the i-th Context. "seq(B,i)" is the i-th byte of the bytestream.
   "seq(b,i)" is the i-th Range coded binary value, "seq(S,0,i)" is the
   i-th initial state, which is 128.  The length of the bytestream
   encoding n binary symbols is "seq(j,n)" bytes.

           seq(r,i) = floor(seq(R,i) * seq(S,i,seq(C,i)) / 2^8)

        seq(S,i+1,seq(C,i)) = seq(zero_state,seq(S,i,seq(C,i))) XOR
        seq(l,i) = seq(L,i) XOR seq(t,i) = seq(R,i) - seq(r,i) <==
        seq(b,i) = 0 <==> seq(L,i) < seq(R,i) - seq(r,i)

   seq(S,i+1,seq(C,i)) = seq(one_state,seq(S,i,seq(C,i))) XOR
   seq(l,i) = seq(L,i) - seq(R,i) + seq(r,i) XOR seq(t,i) = seq(r,i) <==
   seq(b,i) = 1 <==> seq(L,i) >= seq(R,i) - seq(r,i)

                seq(S,i+1,k) = seq(S,i,k) <== seq(C,i) != k

             seq(R,i+1) = 2^8 * seq(t,i) XOR
             seq(L,i+1) = 2^8 * seq(l,i) + seq(B,seq(j,i)) XOR
             seq(j,i+1) = seq(j,i) + 1 <== seq(t,i) < 2^8

                 seq(R,i+1) = seq(t,i) XOR
                 seq(L,i+1) = seq(l,i) XOR
                 seq(j,i+1) = seq(j,i) <== seq(t,i) >= 2^8

                             seq(R,0) = 65280

                   seq(L,0) = 2^8 * seq(B,0) + seq(B,1)

                               seq(j,0) = 2
```

Please review.

dericed:
ping?

retokromer:
I agree with using XOR, which is clear.

I’m afraid, to me the use of `seq()` seems less clear than the use of subscript.

In my opinion, `<==` etc. is indeed an understandable ASCII Art for \Longleftarrow etc.

dericed:
@retokromer, Any suggestion for an ascii equivalent to the use of subscripts?

retokromer:
@dericed In typewriter is was one single step down… I personally use `A_ij`, but I’m aware that this is, of course, heavily LaTeX inspired.

dericed:
going typewriter-style and including the subscript on the next line with indentation is tricky here because the equations already need more than a single wrap, so it would need to be clear what is line wrap and what is subscript. I like the idea of A_ij as it's cleaner, but in this case would lead to things like `one_state_S_i_C_i`.

dericed:
In case anyone prefers this, here's an example of an equation redone with line-wrapping plus single-step-down for subscript:

```

                       S     = zero_state       XOR
                        i+1              S
                           C              i
                            i              C
                                            i

                       l  = L  XOR t  = R  - r  <==
                        i    i      i    i    i

                       b  = 0 <==> L  < R  - r
                        i           i    i    i
```

retokromer:
What about a mix of the two styles?
```
S_i+1    = zero_state       XOR
     C_i             S_i
                        C_i

l_i = L_i XOR t_i = R_i - r_1 <==

b_i = 0 <==> L_1 < R_1 - r_1
```

dericed:
The mix does seem more intuitive but may perhaps be harder to explain. Any other votes for preference?

dericed:
Hi, all I force updated the PR after another round of revisions. For the RFC output (not the non-RFC PDF output) I defined this syntax in the mathematical function section:

```
   a_{b} the b-th value of a sequence of a

   a_{b,c} the 'b,c'-th value of a sequence of a
```

and then used that rather than subscript on multiple lines. I also focused on alignment to make line wrapping more sensible. Examples of the output for the RFC are:

```
                   context =              Q_0[l-tl] +
                             abs(Q_0) * ( Q_1[tl-t] +
                             abs(Q_1) * ( Q_2[t-tr] +
                             abs(Q_2) * ( Q_3[L-l]  +
                             abs(Q_3) *   Q_4[T-t]  )))
```

and

```
   To encode binary digits efficiently a Range coder is used.  "C_{i}"
   is the i-th Context.  "B_{i}" is the i-th byte of the bytestream.
   "b_{i}" is the i-th Range coded binary value, "S_{0,i}" is the i-th
   initial state, which is 128.  The length of the bytestream encoding n
   binary symbols is "j_{n}" bytes.

              r_{i} = floor( ( R_{i} * S_{i,C_{i}} ) / 2^8 )

               S_{i+1,C_{i}} =  zero_state_{S_{i,C_{i}}} XOR
                         l_i =  L_i                     XOR
                         t_i =  R_i - r_i               <==
                         b_i =  0                       <==>
                         L_i <  R_i - r_i

               S_{i+1,C_{i}} =  one_state_{S_{i,C_{i}}}  XOR
                         l_i =  L_i - R_i + r_i         XOR
                         t_i =  r_i                     <==
                         b_i =  1                       <==>
                         L_i >= R_i - r_i

                     S_{i+1,k} = S_{i,k} <== C_i != k

               R_{i+1} =  2^8 * t_{i}                   XOR
               L_{i+1} =  2^8 * l_{i} + B_{j_{i}}       XOR
               j_{i+1} =  j_{i} + 1                     <==
               t_{i}   <  2^8

               R_{i+1} =  t_{i}                         XOR
               L_{i+1} =  l_{i}                         XOR
               j_{i+1} =  j_{i}                         <==
               t_{i}   >= 2^8

                               R_{0} = 65280

                        L_{0} = 2^8 * B_{0} + B_{1}

                                 j_{0} = 2
```

After the prior attempts and getting advice from other RFC authors that have tried to negotiate equations into RFC, I think this is about as good as possible in ASCII. Any advice on the phrasing of the mathematical functions I added, (a_{b})?

retokromer:
LGTM, yet I would personally write
```
context = Q_0[l-tl]
          + abs(Q_0) * ( Q_1[tl-t]
                         + abs(Q_1) * ( Q_2[t-tr]
                                        + abs(Q_2) * ( Q_3[L-l]
                                                       + abs(Q_3) * Q_4[T-t] )))
```
is ASCII art.

dericed:
Your example is 8 characters too wide for an RFC.

retokromer:
```
context = Q_0[l-tl]
        + abs(Q_0) * ( Q_1[tl-t]
                   + abs(Q_1) * ( Q_2[t-tr]
                              + abs(Q_2) * ( Q_3[L-l]
                                         + abs(Q_3) * Q_4[T-t] )))
```
would fit, but is less readable :-(

OK with your version!

michaelni:
patch applied

michaelni:
This makes ffv1.md quite ugly: for example: https://github.com/FFmpeg/FFV1/blob/master/ffv1.md

dericed:
Agreed, though the RFC output looks quite better while the PDF output is the same. Since the PDF and RFC are more official while the md is more like source code, this approach seemed a good compromise to me if we want to simultaneously support both the LaTeX features of PDF and the ascii style of RFC.

dericed:
ah i see more what you mean, that the PDF and RFC versions of the equation are put on the same line in the markdown rendering. I think I can fix that with spacing, will test soon.

PR #52, opened 2017-02-18
move Michael's copyright statement from spec to pdf backmatter

dericed:
since the RFC build already includes a boilerplate IETF copyright
statement:
> Copyright (c) 2017 IETF Trust and the persons identified as the
document authors.  All rights reserved.

### DIFF
--- a/ffv1.md
+++ b/ffv1.md
@@ -1222,7 +1222,7 @@ See <https://github.com/FFmpeg/FFV1/commits/master>
 
 -   mean,k estimation for the Golomb Rice codes
 
-# Copyright
-
-Copyright 2003-2013 Michael Niedermayer \<michaelni@gmx.at\>
-This text can be used under the GNU Free Documentation License or GNU General Public License. See <http://www.gnu.org/licenses/fdl.txt>.
+PDF:# Copyright
+PDF:
+PDF:Copyright 2003-2013 Michael Niedermayer \<michaelni@gmx.at\>
+PDF:This text can be used under the GNU Free Documentation License or GNU General Public License. See <http://www.gnu.org/licenses/fdl.txt>.


michaelni:
With that change ffv1.md has no copyright statement in it anymore

retokromer:
I agree with @michaelni: «something» should be also inside the `ffv1.md` file.

dericed:
The copyright statement is added by mmark. If the equation pr is ok perhaps we should leave the copyright statement here but prefix with "PDF:"

dericed:
This PR is updated according to @michaelni and @retokromer's comments. So that the copyright is still in the markdown, used in the PDF output, but kept out of the RFC output (where it would conflict with the IETF's copyright notice.

I note that this PR uses the same `PDF:` `RFC:` line prefixing strategy that creates ugliness in the rendered markdown, as @michaelni notes in https://github.com/FFmpeg/FFV1/pull/50#issuecomment-293539268. I have an idea to replace `^PDF:` and `RFC:` as line prefixes with potentially a style like:

```
[RFC]
ascii art for RFC
[/RFC]
[PDF]
LaTeX for PDF
[/PDF]
```

but suggest that this potential transition does not block this pull request. Currently when the RFC build it contains two copyright notices that conflictingly claim copyright to both Michael and to the IETF.

Please re-review.

retokromer:
Hmm… Should it not be `2017 IETF` (or both
```
2003-2013 Michael Niedermayer
2017 IETF
```
)? Just asking – Legality is not really my thing.

dericed:
mmark, which translates the markdown to the RFC XML format, inserts this boilerplate:

```
Copyright Notice

   Copyright (c) 2017 IETF Trust and the persons identified as the
   document authors.  All rights reserved.

   This document is subject to BCP 78 and the IETF Trust's Legal
   Provisions Relating to IETF Documents
   (http://trustee.ietf.org/license-info) in effect on the date of
   publication of this document.  Please review these documents
   carefully, as they describe your rights and restrictions with respect
   to this document.  Code Components extracted from this document must
   include Simplified BSD License text as described in Section 4.e of
   the Trust Legal Provisions and are provided without warranty as
   described in the Simplified BSD License.
```

Since Michael is a listed author, he is described here as a joint copyright holder with the IETF Trust and the other listed authors. I'm not certain but would expect that IETF may object if there are other copyright assertions beyond this one.

PR #53, opened 2017-02-18
add self as co-author

dericed:


JeromeMartinez:
so... Me too?

dericed:
If the situation fits the description at http://www.rfc-editor.org/pipermail/rfc-interest/2015-May/008869.html

ablwr:
👍

michaelni:
patches applied

Issue #54, opened 2017-04-11
standardize "color space" vs "colorspace"


dericed:
both are used

ablwr:
Is there a preference?

dericed:
https://tools.ietf.org/html/rfc6386 seems to prefer "color space", I think I would too.

dericed:
@ablwr resolved this in https://github.com/FFmpeg/FFV1/commit/90aea8bdc680a0e47f48faf7e042c0171db9a494.

PR #55, opened 2017-04-12
updates instances of colorspace to "color space" for consistency

ablwr:
Correcting use of "colorspace" (when not a declared variable) for #54


dericed:
+1

michaelni:
patch applied

PR #56, opened 2017-04-14
Improvement of context and quantization tables paragraphs

JeromeMartinez:


ffv1.md
```diff
@@ -201,28 +203,41 @@ Background: a two's complement signed 16-bit signed integer was used for storing
 +---+---+---+---+
 ```
 
-The quantized sample differences L-l, l-tl, tl-t, t-T, t-tr are used as context:
+The quantized sample differences are used as context:
 
-PDF:$$context=Q_{0}[l-tl]+\left|Q_{0}\right|(Q_{1}[tl-t]+\left|Q_{1}\right|(Q_{2}[t-tr]+\left|Q_{2}\right|(Q_{3}[L-l]+\left|Q_{3}\right|Q_{4}[T-t])))$$
+PDF:$$context=Q_{0}[l-tl]+Q_{1}[tl-t]+Q_{2}[t-tr]+Q_{3}[L-l]+Q_{4}[T-t]$$
```
dericed: I had presumed that the pipes (`|a|`) were to expressed absolute values, but in this edit they are simply removed. Does this semantic meaning?

ffv1.md
```diff
@@ -191,6 +191,8 @@ Background: a two's complement signed 16-bit signed integer was used for storing
 
 ## Context
 
+For samples, compared to pixel X, named:
```
dericed: Suggested rewording:

> Samples, positioned relative to pixel `X`, are referenced to with the following abbreviations:

ffv1.md
```diff
@@ -201,28 +203,41 @@ Background: a two's complement signed 16-bit signed integer was used for storing
...
-RFC:          abs(Q_3) *   Q_4[T−t]  )))
+RFC:context = Q_{0}[l − tl] +
+RFC:          Q_{1}[tl − t] +
+RFC:          Q_{2}[t − tr] +
+RFC:          Q_{3}[L − l]  +
+RFC:          Q_{4}[T − t]
 RFC:```
 
-If the context is smaller than 0 then -context is used and the difference between the sample and its predicted value is encoded with a flipped sign.
+If context >= 0 then `context` is used and the difference between the sample and its predicted value is encoded as is, else `-context` is used and the difference between the sample and its predicted value is encoded with a flipped sign.
```
dericed: Consider tilde quotes on `context >= 0`.

Do I understand correctly that if context is equal to `-1` then since `-1` is not greater than 0, then `-(-1)` is used, ie `1`?

ffv1.md
```diff
@@ -201,28 +203,41 @@ Background: a two's complement signed 16-bit signed integer was used for storing
 +---+---+---+---+
 ```
 
-The quantized sample differences L-l, l-tl, tl-t, t-T, t-tr are used as context:
+The quantized sample differences are used as context:
```
dericed: Consider `These 5 quantized sample differences are used as context:`

ffv1.md
JeromeMartinez: There is abs() in the current RFC part, but I don't see what is absolute value of a table, and I don't see corresponding code in FFmpeg source code.
Note that the RFC version was containing multiplication too, and I don't see it too in FFmpeg source code.
See this [snipset](https://github.com/FFmpeg/FFmpeg/blob/74314f1f5f9ef69700eb18b85a8260e2754a31ef/libavcodec/ffv1_template.c#L32-L53).

I am not sure of it, help would be appreciated for confirmation.

ffv1.md
```diff
@@ -201,28 +203,41 @@ Background: a two's complement signed 16-bit signed integer was used for storing
...
+If context >= 0 then `context` is used and the difference between the sample and its predicted value is encoded as is, else `-context` is used and the difference between the sample and its predicted value is encoded with a flipped sign.
 
-## Quantization
+## Quantization tables
 
-There are 5 quantization tables for the 5 sample differences, both the number of quantization steps and their distribution are stored in the bitstream. Each quantization table has exactly 256 entries, and the 8 least significant bits of the sample difference are used as index:
+For each quantization table set, there are 5 quantization tables for the 5 sample differences, both the number of quantization steps and their distribution are stored in the bitstream. Each quantization table has exactly 256 entries, and the 8 least significant bits of the sample difference are used as index:
 
-PDF:$$Q_{i}[a-b]=Table_{i}[(a-b)\&255]$$
+PDF:$$Q_{j}[k]=quant_tables[i][j][k\&255]$$
```
dericed: Meaning of `j` and `k`?

ffv1.md
```diff
@@ -201,28 +203,41 @@ Background: a two's complement signed 16-bit signed integer was used for storing
...
 RFC:```
-RFC:Q_{i}[a − b] = Table_{i}[(a − b)&255]
+RFC:Q_{j}[k] = quant_tables[i][j][k&255]
 RFC:```
 
+In this formula, `i` is the quantization table sets index (see below).
+
+## Quantization table sets
+
+Several quantization table sets can be used in the same bitstream.
```
dericed: Can this be more specific? Is the number of quantization table sets per bitstream: `>=0`, `>=1`?
Is there a max limit to the count of quantization table?

ffv1.md
JeromeMartinez: Right. Contexts can be positive only, the sign of stored context seems to be used only for saying that the encoded difference must be subtracted from the predicted value instead of added.
See this [snipset](https://github.com/FFmpeg/FFmpeg/blob/b1d12509384b9285b85f5c183772d75a81e74009/libavcodec/ffv1dec_template.c#L51-L55)

ffv1.md
```diff
@@ -201,28 +203,41 @@ Background: a two's complement signed 16-bit signed integer was used for storing
...
+
+For each plane of each slice, a quantization table set is selected from an index:
+- For Y plane, `quant_table_index [ 0 ]` index is used
+- For Cb and Cr planes, `quant_table_index [ 1 ]` index is used
+- For Alpha plane, `quant_table_index [ (version <= 3 || chroma_planes) ? 2 : 1 ]` index is used
+
+Background: in first implementations of FFV1 bitstream, the index for Cb and Cr planes was stored even if it is not used (chroma_planes set to 0), this index is kept for version <= 3 on order to keep compatibility with bitstreams in the wild.
```
dericed: `on order` -> `in order`

ffv1.md
```diff
@@ -1105,37 +1120,35 @@ initial\_state[ i ][ j ][ k ] = ( pred + initial\_state\_delta[ i ][ j ][ k ] )
...
-
 ```c
 function                                                      | type
 --------------------------------------------------------------|-----
 QuantizationTablePerContext(i, j, scale) {                    |
     v = 0                                                     |
     for( k = 0; k < 128; ) {                                  |
-        len - 1                                               | sr
-        for( a = 0; a < len; a++ ) {                          |
+        len_minus1                                            | ur
```
dericed: Consider defining `len_minus` as a temporary counter.

ffv1.md
JeromeMartinez: they are arbitrary values in that case, as `j` and `k` are on the left and right, and they'll be in formulas when `Q` is used.
In practice, `j` is the context and k the pixel difference.
I don't think it is need to define them here, as we see that when `Q` is used.

ffv1.md
JeromeMartinez: `>=1`, it is `quant_table_count` value. maybe we can add the constraint "must not be 0" in `quant_table_count` definition.

FFmpeg has an hard-coded maximum of 8, but looks like this is not necessary in specs, so "unlimited" (as much as width or height i.e. at some point it is too much :) ).

ffv1.md
dericed: I added the `abs()` in the RFC intended it to be a semantic equivalent to the pipes used in the LaTeX version. In the context of the plain text RFC, I thought `abs()` would be more readable than `|a|`. But I am not certain that the pipes in the LaTeX version indicate an absolute value but that is what the mathematical function section implies. Ping to @michaelni.

ffv1.md
JeromeMartinez: it is not a counter, it is a temporary value, so not defined. It is a bit explained in the intro ("storing the number of equal entries -1"), does it need a dedicated explicit definition? as it is temporary, I suggest not to do it.

JeromeMartinez:
PR updated from comments from Dave, and including wording from https://github.com/FFmpeg/FFV1/pull/64 (typo fix is not backported as the list is completely removed, in order to avoid duplicated and potentially unsynchronized content)

michaelni:
This contains quite a mixture of changes, some purely cosmetic some bitstream changes. These belong to several seperate commits

ffv1.md
```diff
@@ -218,28 +220,41 @@ Background: a two's complement signed 16-bit signed integer was used for storing
...
+RFC:          Q_{2}[t − tr] +
+RFC:          Q_{3}[L − l]  +
+RFC:          Q_{4}[T − t]
 RFC:```
 
-If the context is smaller than 0 then -context is used and the difference between the sample and its predicted value is encoded with a flipped sign.
+If `context >= 0` then `context` is used and the difference between the sample and its predicted value is encoded as is, else `-context` is used and the difference between the sample and its predicted value is encoded with a flipped sign.
 
-## Quantization
+## Quantization tables
```
dericed: This section header is already used on line 1174. Duplicate section names should be avoided.

ffv1.md
```diff
@@ -218,28 +220,41 @@ Background: a two's complement signed 16-bit signed integer was used for storing
 +---+---+---+---+
 ```
 
-The quantized sample differences L-l, l-tl, tl-t, t-T, t-tr are used as context:
+Relative to any sample `X`, the 5 quantized sample differences are used as context:
```
dericed: suggesting: `the following 5 quantized`

ffv1.md
retokromer: a positive integer

JeromeMartinez:
> These belong to several seperate commits

Got it. Let's start with https://github.com/FFmpeg/FFV1/pull/68 then I'll update this PR.

dericed:
status?

JeromeMartinez:
PR updated, @michaelni and @dericed please review again.

I don't see multiplications, `abs` and parenthesis at this place in FFmpeg FFV1 decoder code, so I remove them from specs.
I understand the complex computing for context is already done by a pre-computing in `QuantizationTable(i, j, scale)` function, with multiplication of intermediate `scale` value, keeping multiplications, `abs` and parenthesis at this place would make the compute used twice instead of one.

(or... What do I miss in the  FFmpeg FFV1 decoder code?)

michaelni:
> LGTM, but admit that I don't understand this equation well. I think I had misinterpreted the LaTeX when I added the parenthesis and abs.

Yes, i must have missed it but the |Q| stuff was meant as the number of elements / magnitude in the quantization table not as the absolute value of something

Issue #35, opened 2017-01-03
explain convention of documenting component structures


dericed:
In https://github.com/FFmpeg/FFV1/blob/master/ffv1.md#conventions it references the `C standard` for operators and order of precedence, but `C` is also used to document structures within FFV1, such as in https://github.com/FFmpeg/FFV1/blob/master/ffv1.md#configuration-record, https://github.com/FFmpeg/FFV1/blob/master/ffv1.md#frame, and https://github.com/FFmpeg/FFV1/blob/master/ffv1.md#slice. I think it could be made more clear that these sections are in `C` by referencing this usage in the Conventions section.

JeromeMartinez:
we use "if", "else" and so on as in C standard, but the structure is not C, so we need a dedicated paragraph about it.

dericed:
@JeromeMartinez can you draft such a paragraph?

dericed:
I found some IETF documentation that pertains to this issue at https://www.ietf.org/iesg/statement/pseudocode-guidelines.html (see references to `pseudocode`).

dericed:
closed via https://github.com/FFmpeg/FFV1/commit/68f6d1e2743df5f8037d4dfa724d44281272656a

PR #62, opened 2017-05-17
describe pseudo-code

dericed:
This is based upon IETF guidelines found at
https://www.ietf.org/iesg/statement/pseudocode-guidelines.html and
consider pseudo-code phrasing from documents such as rfc4271 and
rfc4616.


dericed:
intended to resolve https://github.com/FFmpeg/FFV1/issues/35

PR #63, opened 2017-05-17
update readme to link ietf versions of ffv1

dericed:


dericed:
ping

michaelni:
The new text says "This repository manages the development of specification documents for FFV1, a lossless intra-frame video codec, by the [cellar working group]". This seems incorrect, cellar has not created FFV1. cellar (you and others more precissely) have done huge amounts of cleanup and corrections in the text. I was also hoping that we all would do alot of development and testing of new ideas and features but none of this occured yet. That would make FFV1 more a cellar working group creation.

dericed:
ah, the ordering was a bit awkward, I had meant it to be understood that it was the `development of specification documents` which was `by the cellar working group`, not to imply that FFV1 as a video encoding format was by cellar. Will reword and resubmit.

dericed:
The PR is updated. Please re-review.

retokromer:
I share @michaelni’s sense of frustration. Since one year it’s always the wrong timing for new ideas and developments.

JeromeMartinez:
I think we all share the idea of having new features;
About the testing, I kindly disagree as we have an independent additional decoder and I added a couple of exceptions in the spec especially because the stream was tested, actually.
Before new ideas, we need a clear spec of what already exists.

We take any help for having the spec accepted by IETF.

PR #66, opened 2017-05-23
Definitions

dericed:


ffv1.md
```diff
@@ -737,7 +739,7 @@ Slice( ) {                                                    |
         SliceHeader( )                                        |
     SliceContent( )                                           |
     if (coder_type == 0)                                      |
-        while (!byte_aligned())                               |
+        while (!byte_aligned( NumBytes ))                     |
```
JeromeMartinez: In that case, `Slice()` should be `Slice ( NumBytes )` (2 places).

I am a bit reluctant to put `NumBytes` everywhere, it make the read a bit more complex without real value added IMHO, I would remove it from `while ( remaining_bits_in_bitstream( NumBytes ) )`

Someone else, opinion?

dericed:
I agree with @JeromeMartinez's comment at https://github.com/FFmpeg/FFV1/pull/66#discussion_r118030766 and dropped the `NumBytes` patch. The remaining commits should be easier to gain consensus on. Please re-review.

PR #68, opened 2017-06-03
Make Quantization descriptions more coherent

JeromeMartinez:
I understand that there is some incoherencies in the description of Quantization related items, i.e. "context" is used with 2 different meanings, the are in some places "quantization table set" never defined and in other places "quantization table" which is similar the "quantization table set".

I tried to have something more coherent, with 3 items:
- "quantization tables", having quant_table_count "quantization table"
- "quantization table", having 5 "blocks"
- "block", having 256 entries

At the beginning, I tried to do with:
- "quantization tables sets", having quant_table_set_count "quantization table set"
- "quantization table set", having 5 "quantization table"
- "quantization table", having 256 entries
but it was more changes in the document, and I preferred to avoid so many changes, is it OK?

Other changes:
`a − b` replaced by `k`, as `a − b` detail is not needed at this place (and `k` is chosen because it is already used in QuantizationTable() definition)
`set` removed when present, in order to have a single item for the same thing
`Table_` is removed because it is never defined and same as quant_tables[i]
`Quantization table index` is described (with more background about the exception)
`QuantizationTablePerContext` changed to `QuantizationTablePerBlock` because `Context` is also used (see Context paragraph)
Added missing space in the list of negative numbers

### DIFF
--- a/ffv1.md
+++ b/ffv1.md
@@ -240,7 +240,7 @@ Background: a two's complement signed 16-bit signed integer was used for storing
 
 ## Context
 
-Relative to any sample `X`, as represented in [the Template for Context and Median Prediction](#template), the quantized sample differences `L-l`, `l-tl`, `tl-t`, ` T-t`, and `t-tr` are used as context:
+Relative to any sample `X`, as represented in [the Template for Context and Median Prediction](#template), the Quantized Sample Differences `L-l`, `l-tl`, `tl-t`, ` T-t`, and `t-tr` are used as context:
 
 PDF:$$context=Q_{0}[l-tl]+\left|Q_{0}\right|(Q_{1}[tl-t]+\left|Q_{1}\right|(Q_{2}[t-tr]+\left|Q_{2}\right|(Q_{3}[L-l]+\left|Q_{3}\right|Q_{4}[T-t])))$$
 RFC:```
@@ -253,14 +253,28 @@ RFC:```
 
 If the context is smaller than 0 then -context is used and the difference between the sample and its predicted value is encoded with a flipped sign.
 
-## Quantization
+## Quantization Table Sets
 
-There are 5 quantization tables for the 5 sample differences, both the number of quantization steps and their distribution are stored in the bitstream. Each quantization table has exactly 256 entries, and the 8 least significant bits of the sample difference are used as index:
+The bitstream contains 1 or more Quantization Table Sets.  
+Each Quantization Table Set contains exactly 5 Quantization Tables, each Quantization Table corresponding to 1 of the 5 Quantized Sample Differences.  
+For each Quantization Table, both the number of quantization steps and their distribution are stored in the bitstream; each Quantization Table has exactly 256 entries, and the 8 least significant bits of the Quantized Sample Difference are used as index:
 
-PDF:$$Q_{i}[a-b]=Table_{i}[(a-b)\&255]$$
+PDF:$$Q_{j}[k]=quant_tables[i][j][k\&255]$$
 RFC:```
-RFC:Q_{i}[a − b] = Table_{i}[(a − b)&255]
+RFC:Q_{j}[k] = quant_tables[i][j][k&255]
 RFC:```
+
+In this formula, `i` is the Quantization Table Set index, `j` is the Quantized Table index, `k` the Quantized Sample Difference.
+
+## Quantization Table Set indexes
+
+For each plane of each slice, a Quantization Table Set is selected from an index:
+
+- For Y plane, `quant_table_set_index [ 0 ]` index is used
+- For Cb and Cr planes, `quant_table_set_index [ 1 ]` index is used
+- For Alpha plane, `quant_table_set_index [ (version <= 3 || chroma_planes) ? 2 : 1 ]` index is used
+
+Background: in first implementations of FFV1 bitstream, the index for Cb and Cr planes was stored even if it is not used (chroma_planes set to 0), this index is kept for version <= 3 in order to keep compatibility with bitstreams in the wild.
 
 ## Color space
 
@@ -371,9 +385,9 @@ In JPEG2000-RCT color space, the coding order would be left to right and then to
 
 Y[1,1] Y[2,1] Cb[1,1] Cb[2,1] Cr[1,1] Cr[2,1] Y[1,2] Y[2,2] Cb[1,2] Cb[2,2] Cr[1,2] Cr[2,2]
 
-## Coding of the sample difference
+## Coding of the Sample Difference
 
-Instead of coding the n+1 bits of the sample difference with Huffman or Range coding (or n+2 bits, in the case of RCT), only the n (or n+1) least significant bits are used, since this is sufficient to recover the original sample. In the equation below, the term "bits" represents bits_per_raw_sample+1 for RCT or bits_per_raw_sample otherwise:
+Instead of coding the n+1 bits of the Sample Difference with Huffman or Range coding (or n+2 bits, in the case of RCT), only the n (or n+1) least significant bits are used, since this is sufficient to recover the original sample. In the equation below, the term "bits" represents bits_per_raw_sample+1 for RCT or bits_per_raw_sample otherwise:
 
 PDF:$$coder\_input=\left[\left(sample\_difference+2^{bits-1}\right)\&\left(2^{bits}-1\right)\right]-2^{bits-1}$$
 RFC:```
@@ -779,8 +793,8 @@ SliceHeader( ) {                                              |
     slice_y                                                   | ur
     slice_width - 1                                           | ur
     slice_height - 1                                          | ur
-    for( i = 0; i < quant_table_index_count; i++ )            |
-        quant_table_index [ i ]                               | ur
+    for( i = 0; i < quant_table_set_index_count; i++ )        |
+        quant_table_set_index [ i ]                           | ur
     picture_structure                                         | ur
     sar_num                                                   | ur
     sar_den                                                   | ur
@@ -811,13 +825,13 @@ Inferred to be 1 if not present.
 `slice_height` indicates the height on the slice raster formed by num_v_slices.
 Inferred to be 1 if not present.
 
-### quant_table_index_count
+### quant_table_set_index_count
 
-`quant_table_index_count` is defined as 1 + ( ( chroma_planes || version \<= 3 ) ? 1 : 0 ) + ( alpha_plane ? 1 : 0 ).
+`quant_table_set_index_count` is defined as 1 + ( ( chroma_planes || version \<= 3 ) ? 1 : 0 ) + ( alpha_plane ? 1 : 0 ).
 
-### quant_table_index
+### quant_table_set_index
 
-`quant_table_index` indicates the index to select the quantization table set and the initial states for the slice.
+`quant_table_set_index` indicates the Quantization Table Set index to select the Quantization Table Set and the initial states for the slice.
 Inferred to be 0 if not present.
 
 ### picture_structure
@@ -998,12 +1012,12 @@ Parameters( ) {                                               |
     if (version >= 3) {                                       |
         num_h_slices - 1                                      | ur
         num_v_slices - 1                                      | ur
-        quant_table_count                                     | ur
+        quant_table_set_count                                 | ur
     }                                                         |
-    for( i = 0; i < quant_table_count; i++ )                  |
-        QuantizationTable( i )                                |
+    for( i = 0; i < quant_table_set_count; i++ )              |
+        QuantizationTableSet( i )                             |
     if (version >= 3) {                                       |
-        for( i = 0; i < quant_table_count; i++ ) {            |
+        for( i = 0; i < quant_table_set_count; i++ ) {        |
             states_coded                                      | br
             if (states_coded)                                 |
                 for( j = 0; j < context_count[ i ]; j++ )     |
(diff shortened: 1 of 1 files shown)


ffv1.md
```diff
@@ -233,15 +233,28 @@ RFC:```
 
 If the context is smaller than 0 then -context is used and the difference between the sample and its predicted value is encoded with a flipped sign.
 
-## Quantization
+## Quantization tables
 
-There are 5 quantization tables for the 5 sample differences, both the number of quantization steps and their distribution are stored in the bitstream. Each quantization table has exactly 256 entries, and the 8 least significant bits of the sample difference are used as index:
+Each quantization table contains 5 blocks, each block corresponding to 1 of the 5 sample differences, both the number of quantization steps and their distribution are stored in the bitstream. Each quantization table block has exactly 256 entries, and the 8 least significant bits of the sample difference are used as index:
```
dericed: Perhaps clearer language. For instance a table of 10 blocks does `contain 5 blocks`. Could you say `contains exactly 5 blocks`?

dericed: Perhaps a titlecase, backtick-quoted term `Quantized Sample Differences` should be used both in this section and in the prior Context section to clarify that these are the same thing.

dericed: Similar to prior comment the concept of `blocks` and `quantization table block` above could be a more formal term such as:

> Each `Quantization Table` contains 5 `Quantization Table Blocks` with each block corresponding to 1 of the 5 `Quantized Sample Differences`. Both the number of quantization steps and their distribution are stored in the bitstream. Each `Quantization Table Block` has exactly 256 entries, and the 8 least significant bits of the `Quantized Sample Difference` used as index:

dericed: Regarding `each block corresponding to 1 of the 5 sample differences`, is an order implied?

dericed: Regarding `are stored in the bitstream`, could this be more specific as to where this is stored?

ffv1.md
```diff
@@ -233,15 +233,28 @@ RFC:```
...
 
-PDF:$$Q_{i}[a-b]=Table_{i}[(a-b)\&255]$$
+PDF:$$Q_{j}[k]=quant_tables[i][j][k\&255]$$
 RFC:```
-RFC:Q_{i}[a − b] = Table_{i}[(a − b)&255]
+RFC:Q_{j}[k] = quant_tables[i][j][k&255]
 RFC:```
 
+In this formula, `i` is the quantization table index of the currently parsed data, `j` is the sample difference index, `k` the sample difference.
```
dericed: Do I understand correctly, that the `=` would be read as `is assigned`?

ffv1.md
```diff
@@ -233,15 +233,28 @@ RFC:```
...
+
+## Quantization table indexes
+
+Several quantization tables can be used in the same bitstream.
```
dericed: Is this sentence needed? It just seems to say that the bitstream (ffv1 bitstream, I suppose) can contain more than 1 quantization table.

ffv1.md
```diff
@@ -797,7 +810,7 @@ Inferred to be 1 if not present.
 
 ### quant_table_index
 
-`quant_table_index` indicates the index to select the quantization table set and the initial states for the slice.
+`quant_table_index` indicates the index to select the quantization table and the initial states for the slice.
```
dericed: Since `Quantization table indexes` is designed earlier, I think the term should be used in the definition here.

ffv1.md
JeromeMartinez: Except `each block corresponding to 1 of the 5 sample differences` comments (IMO order is explicated with `Q_1`, `Q_2`... And will be more explicit with the next PR changing them in an array), all comments are included in a the new version.

ffv1.md
JeromeMartinez: Right (see "Assignment operators" part)

ffv1.md
JeromeMartinez: Removed

JeromeMartinez:
I added another commit so changes compared to remarks are visible, please review then I can squash them.

ffv1.md
```diff
@@ -233,15 +233,26 @@ RFC:```
 
 If the context is smaller than 0 then -context is used and the difference between the sample and its predicted value is encoded with a flipped sign.
 
-## Quantization
+## Quantization tables
 
-There are 5 quantization tables for the 5 sample differences, both the number of quantization steps and their distribution are stored in the bitstream. Each quantization table has exactly 256 entries, and the 8 least significant bits of the sample difference are used as index:
+Each Quantization Table contains 5 exactly Quantization Table Blocks, each block corresponding to 1 of the 5 Quantized Sample Differences, both the number of quantization steps and their distribution [are stored in the bitstream](#quantization-table). Each Quantization Table Block has exactly 256 entries, and the 8 least significant bits of the Quantized Sample Differences are used as index:
```
dericed: I think you mean `exactly 5` rather than `5 exactly`.

dericed: I don't understand the intent of `and their distribution [are stored in the bitstream](#quantization-table)`. The bracketed label does not seem to match the section link.

ffv1.md
```diff
@@ -233,15 +233,26 @@ RFC:```
...
 RFC:```
-RFC:Q_{i}[a − b] = Table_{i}[(a − b)&255]
+RFC:Q_{j}[k] = quant_tables[i][j][k&255]
 RFC:```
 
+In this formula, `i` is the Quantization Table index of the currently parsed data, `j` is the Quantized Sample Differences index, `k` the Quantized Sample Differences.
+
+## Quantization Table indexes
+
+For each plane of each slice, a Quantization Table is selected from an index:
```
dericed: You may need an extra line break after line 249 for the list to render correctly in the RFC outputs.

ffv1.md
```diff
@@ -160,7 +160,7 @@ NumBytes is a non-negative integer that expresses the size in 8-bit octets of pa
 
 # General Description
 
-Samples within a plane are coded in raster scan order (left->right, top->bottom). Each sample is predicted by the median predictor from samples in the same plane and the difference is stored see [Coding of the Sample Difference](#coding-of-the-sample-difference).
+Samples within a plane are coded in raster scan order (left->right, top->bottom). Each sample is predicted by the median predictor from samples in the same plane and the difference is stored see [Coding of the Quantized Sample Differences](#coding-of-the-quantized-sample-difference).
```
dericed: There's a mismatch in the section link.
```
coding-of-the-quantized-sample-difference
vs
Coding of the Quantized Sample Differences
```

I suggest adding an `s` to the end of the link.

JeromeMartinez:
@dericed I added another update, please review.

I did not change something for:

> I don't understand the intent of and their distribution [are stored in the bitstream](#quantization-table). The bracketed label does not seem to match the section link.

as I don't see what you expect: you wanted to know where the quantization tables are stored, are linked to the corresponding place ("Quantization table" section in the bitstream part of the specifications).
Please suggest another method if this one does not fit.

JeromeMartinez:
Commits are combined now, please review again and confirm all is fine.

ffv1.md
```diff
@@ -160,7 +160,7 @@ NumBytes is a non-negative integer that expresses the size in 8-bit octets of pa
 
 # General Description
 
-Samples within a plane are coded in raster scan order (left->right, top->bottom). Each sample is predicted by the median predictor from samples in the same plane and the difference is stored see [Coding of the Sample Difference](#coding-of-the-sample-difference).
+Samples within a plane are coded in raster scan order (left->right, top->bottom). Each sample is predicted by the median predictor from samples in the same plane and the difference is stored see [Coding of the Quantized Sample Differences](#coding-of-the-quantized-sample-differences).
```
michaelni: The sample difference is coded not the quantized sample difference.

ffv1.md
```diff
@@ -233,15 +233,27 @@ RFC:```
 
 If the context is smaller than 0 then -context is used and the difference between the sample and its predicted value is encoded with a flipped sign.
 
-## Quantization
+## Quantization tables
 
-There are 5 quantization tables for the 5 sample differences, both the number of quantization steps and their distribution are stored in the bitstream. Each quantization table has exactly 256 entries, and the 8 least significant bits of the sample difference are used as index:
+Each Quantization Table contains exactly 5 Quantization Table Blocks, each block corresponding to 1 of the 5 Quantized Sample Differences, both the number of quantization steps and their distribution are stored in the bitstream. Each Quantization Table Block has exactly 256 entries, and the 8 least significant bits of the Quantized Sample Differences are used as index:
```
michaelni: A Table in the most general sense maps an input parameter to an output parameter.
Before this change each of the 5 tables mapped a bounded scalar integer to a scalar integer of smaller range. Aka performing scalar quantization.
What does the single table do afterwards ? Its input has to be a 5 dimensional vector to fill the same role. While this works it is not how it can be implemented. I think the text was clearer before

dericed: I think I also prefer the original use of the term 'table' as a map rather than introduce the new concept of 'table block' which is a term I have less understanding of. Relatedly I see ISO-14495 uses the term 'mapping table' and 'mapping table entries'. Is it worthwhile sharing this terminology?

JeromeMartinez: Current spec uses at one place "quantization table set", please comment about using this terminology everywhere (it was a request in my PR introduction).

dericed: I am okay with "quantization table set" but it should be defined and probably reused in the line 238 we're discussion above.

JeromeMartinez: @michaelni and @dericed are you both OK for using the alternative form proposed in the intro of my PR ("quantization tables sets", "quantization table set", "quantization table"), meaning that there will be lot of changes a bit everywhere in the spec ("quantization table" must be change nearly everywhere by "quantization table set").
I don't want to spend time on that if the idea is not agreed by everyone first.

dericed: I see above you had:

> At the beginning, I tried to do with:
>    "quantization tables sets", having quant_table_set_count "quantization table set"
>    "quantization table set", having 5 "quantization table"
>    "quantization table", having 256 entries
>    but it was more changes in the document, and I preferred to avoid so many changes, is it > OK?

I don't have a strong opinion, but to me 'quantization table set' and 'quantization table' make sense and I suggest in addition to putting these in backtick quotes and title case that they be defined as local terminology as well.

To me 'quantization tables sets' is less elegant, particularly since it contains data that is not only 'quantization tables set'.

I suggest the 'entries' mentioned above be a defined term as well.

JeromeMartinez: ping, I wish to know in which direction you want me to go in order to get the PR accepted.

michaelni: > I don't have a strong opinion, but to me 'quantization table set' and 'quantization table' make sense [..]
> To me 'quantization tables sets' is less elegant, particularly since it contains data that is not only 'quantization tables set'.

+1

dericed:
Cut-off date for draft updates for IETF99 is July 3rd. Any chance we can resolve the issues of this pull request before then.

JeromeMartinez:
> Any chance we can resolve the issues of this pull request before then.

I plan to work again on it tomorrow and send a new proposal, then it will depends on comments, I guess.

JeromeMartinez:
@dericed and @michaelni:
I switched to the second possibility for renaming.
I also reverted the change about sample difference.

Please review again.

ffv1.md
```diff
@@ -1227,11 +1242,11 @@ QuantizationTablePerContext(i, j, scale) {                    |
 
 ### quant_tables
 
-`quant_tables` indicates the quantification table values.
+`quant_tables[ i ][ j ][ k ]` indicates the quantification table value of the Quantized Sample Difference `k` of the Quantization Table `j` of the Set Quantization Table `i`.
```
dericed: Shouldn't this end with?
> of the Quantization Table Set `i`.

ffv1.md
```diff
@@ -253,15 +253,29 @@ RFC:```
...
-PDF:$$Q_{i}[a-b]=Table_{i}[(a-b)\&255]$$
+PDF:$$Q_{j}[k]=quant_tables[i][j][k\&255]$$
 RFC:```
-RFC:Q_{i}[a − b] = Table_{i}[(a − b)&255]
+RFC:Q_{j}[k] = quant_tables[i][j][k&255]
 RFC:```
 
+In this formula, `i` is the Quantization Table Set index, `j` is the Quantized Table index, `k` the Quantized Sample Difference.
+
+## Quantization Table Set indexes
```
dericed: Nitpick but I would suggest using the singular `index` in the section header.

dericed:
Two minor comments, but otherwise this looks good. Thanks.

ffv1.md
JeromeMartinez: > Shouldn't this end with?
> > of the Quantization Table Set `i`.

Fixed.

ffv1.md
JeromeMartinez: IMO this is a general section about all Quantization Table Set indexes (there are several indexes in each slice, ~1 per plane), I am in favor of keeping plural without being against singular, waiting for another person for suggestion.

michaelni:
This requires to be re checked, there are issues in it, for example quant_tables is interpreted as quant subscript t. There is also trailing whitespace, The "the Template for Context and Median Prediction" link also doesnt work, thats seems to be an older issue though

michaelni:
The "Restrictions" section seems to have "Quantization Table Set" as parent in the pdf at least

JeromeMartinez:
> There is also trailing whitespace

The trailing whitespaces I added are for carriage return when converted to HTML (markdown considers single carriage returns without white space as no carriage return), was from my point of view easier to read. I suggest to keep them, without being against removing them.

I push soon another PR for the other issues.

michaelni:
> The trailing whitespaces I added are for carriage return when converted to HTML (markdown considers single carriage returns without white space as no carriage return), was from my point of view easier to read.

I did not realize that, thx for explaining

JeromeMartinez:
@michaelni 

> quant_tables is interpreted as quant subscript t

I don't understand this sentence.

michaelni:
> > quant_tables is interpreted as quant subscript t

> I don't understand this sentence.

In the pdf output some quant_tables are written with a small t, as if t was a subscript

PR #70, opened 2017-06-07
expanding on median predictor section

dericed:


dericed:
This moves the context-template into the median prediction section. In the RFC version it auto-adds a `Figure #` that use the #anchor to make cross-references to, but it's hard to emulate this behavior exactly in the PDF version. Still I think it makes sense for both the median and context sections to use the same figure and relative-sample codes (tl, l, t, etc). I also defined a median function.

Comments?

ffv1.md
```diff
@@ -111,6 +111,8 @@ log2(a)               the base-two logarithm of a
 
 min(a,b)              the smallest of two values a and b
 
+median(a,b,c)         the numerical middle value in a data set of a, b, and c
```
michaelni: You can define median() as a+b+c-min(a,b,c)-max(a,b,c) if you want a more formal definiton without "middle"

dericed: updated to a mix of the two:

> the numerical middle value in a data set of a, b, and c, i.e. a+b+c-min(a,b,c)-max(a,b,c)

dericed:
PR updated according to @michaelni's feedback. Please re-review.

PR #71, opened 2017-06-07
add iana considerations section

dericed:


JeromeMartinez:
I think we need some IANA registration, or at least some codec info for IETF formats, I added a comment on CELLAR list: https://www.ietf.org/mail-archive/web/cellar/current/msg01168.html

dericed:
updated for a new review

ffv1.md
```diff
@@ -1292,6 +1292,63 @@ The reference implementation [@REFIMPL] was validated in the following condition
...
+
+Subtype name:  FFV1
+
+Required parameters:  None.
+
+Optional parameters:
+
+  This parameter is used to signal the capabilities of a receiver implementation. This parameter MUST NOT be used for any other purpose.
+
+  version:  The version of the FFV1 encoding as defined by [in the section on `version`](#version).
```
JeromeMartinez: I would put version\[.micro_version\].

Similar to https://www.iana.org/assignments/media-types/video/H264 parameters, I would add coder_type, color_space and bits_per_raw_sample, as some decoders may support only some combinations or we may add new coder_type, color_space in the future (like H.264 decoders support only some profiles and/or levels and standard adds new profiles later). Adding chroma subsampling info and others is maybe too much?

Reading IANA consideration for H264, max-fs (frame size), max-br (bit rate), something about max count of slices, could be taken into consideration.

dericed: the pr is updated to consider this comment. please re-review.

ffv1.md
```diff
@@ -1292,6 +1292,71 @@ The reference implementation [@REFIMPL] was validated in the following condition
...
+
+Additional information:  None.
+
+Person & email address to contact for further information:  TODO [Pick a contact]
+
+Intended usage:  COMMON
+
+Restrictions on usage:  None.
+
+Author:  TODO [Pick a contact]
```
dericed: @michaelni, may I set you as the IANA contact?

ffv1.md
JeromeMartinez: Fine for me but willing another pair of eyes on it, I am not confident enough that I saw everything.

dericed: thx, I suspect we'll get a better review when reviewed by the area director or iesg, but for now I hope we have consensus amongst participating contributors/reviewers.

ffv1.md
michaelni: @dericed, what does the IANA contact has to do ?

dericed: Hi @michaelni, the definition of the contact in the media type registration from [rfc4855](https://tools.ietf.org/html/rfc4855) is only
> Person (including contact information) to contact for further information.

IIUC this contact only applies to the media type registration of IANA, but if IANA has some questions during the registration they need a specific contact. If you'd rather not, I could volunteer.

michaelni: ohh, ok, feel free to add me

dericed: @michaelni I added you as contact and myself as the author of the registration section.

dericed:
Updated to add contact names, please re-review.

michaelni:
this breaks build, please test build before approving changes

 draft-ietf-cellar-ffv1-01.xml: Line 1490: IDREF attribute target references an unknown ID "color_space"
 draft-ietf-cellar-ffv1-01.xml: Line 1488: IDREF attribute target references an unknown ID "coder_type"
 draft-ietf-cellar-ffv1-01.xml: Line 1492: IDREF attribute target references an unknown ID "bits_per_raw_sample"
 draft-ietf-cellar-ffv1-01.xml: Line 1486: IDREF attribute target references an unknown ID "micro_version"

michaelni:
It seems the _ are lost in the anchors, i wonder how that happens

PR #77, opened 2017-07-02
White space for forcing carriage return in a paragraph

JeromeMartinez:
When I filled the spec with carriage returns in a paragraph with Markdown, my goal was to output a carriage return in HTML. Actually not the case.
This commit fixes that, adding carriage returns.
Reference for how to put a carriage return in Markdown: https://markdown-guide.readthedocs.io/en/latest/basics.html#line-return

Example of issue: all on the same line on https://github.com/FFmpeg/FFV1/blob/master/ffv1.md#quant_table_set_count, split in 3 lines on https://github.com/MediaArea/FFV1/blob/Markdown/ffv1.md#quant_table_set_count.

### DIFF
--- a/ffv1.md
+++ b/ffv1.md
@@ -721,14 +721,14 @@ ConfigurationRecord( NumBytes ) {                             |
 
 ### reserved_for_future_use
 
-`reserved_for_future_use` has semantics that are reserved for future use.
-Encoders conforming to this version of this specification SHALL NOT write this value.
+`reserved_for_future_use` has semantics that are reserved for future use.  
+Encoders conforming to this version of this specification SHALL NOT write this value.  
 Decoders conforming to this version of this specification SHALL ignore its value.
 
 ### configuration_record_crc_parity
 
-`configuration_record_crc_parity` 32 bits that are chosen so that the `Configuration Record` as a whole has a crc remainder of 0.
-This is equivalent to storing the crc remainder in the 32-bit parity.
+`configuration_record_crc_parity` 32 bits that are chosen so that the `Configuration Record` as a whole has a crc remainder of 0.  
+This is equivalent to storing the crc remainder in the 32-bit parity.  
 The CRC generator polynomial used is the standard IEEE CRC polynomial (0x104C11DB7) with initial value 0.
 
 ### Mapping FFV1 into Containers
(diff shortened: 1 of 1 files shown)


michaelni:
Why do you want to force new lines in the middle of paragraphs ? depending on how much width there is, that may be sub optimal

JeromeMartinez:
I think that:

> quant_table_set_count indicates the number of Quantization Table Sets. Inferred to be 1 if not present. MUST NOT be 0.

is less readable as:

> `quant_table_set_count` indicates the number of Quantization Table Sets.
Inferred to be 1 if not present.
MUST NOT be 0.

As each line is a different part of the definition. Having a different paragraph for each line would be too much.

another problematic example is:
https://github.com/FFmpeg/FFV1/blob/master/ffv1.md#initial_state_delta

> `initial_state_delta` [ i ][ j ][ k ] indicates the initial Range coder state, it is encoded using k as context index and pred = j ? initial_states[ i ][j - 1][ k ] : 128 initial_state[ i ][ j ][ k ] = ( pred + initial_state_delta[ i ][ j ][ k ] ) & 255

It is very difficult to read (for me: it is impossible to understand, I just see a single formula with two `=` and I don't understand the link between each part), with `pred` definition and `initial_state[ i ][ j ][ k ]` definition on the same line, without any separator, but the source is:

> `initial_state_delta` [ i ][ j ][ k ] indicates the initial Range coder state, it is encoded using k as context index and  
pred = j ? initial\_states[ i ][j - 1][ k ] : 128  
initial\_state[ i ][ j ][ k ] = ( pred + initial\_state\_delta[ i ][ j ][ k ] ) & 255

From my point of view the current version (all on 1 line, same separator between a part of the formula and another formula, i.e. space) is not understandable as a definition of of `pred` then a definition of `initial_state`.

This is definitely very subjective, we can abandon this change if nobody else thinks it is useful. The opposite idea would be to remove the carriage returns in the Markdown source as the Markdown version makes thinking that there is a carriage return where there is no carriage return.

Maybe at least keeping modification for formulas (`pred` / `initial_state` formulas issue)?

michaelni:
> From my point of view the current version (all on 1 line, same separator between a part of the formula and another formula, i.e. space) is not understandable as a definition of of pred then a definition of initial_state.

The formular in your example a bit slow to parse in both cases. There are 4 things the text conveys to the reader, 1. what it is, 2. what is stored (delta mod 8 from some predictor), 3. the predictor, 4. the context used for the coder. I agree the newline makes it better in this case but it still intermingles things and isnt very clear

dericed:
Status of this? It appears that @JeromeMartinez's addition of spacing for line breaks makes it more readable. @michaelni points out that it could be more readable, but that could occur in a subsequent commit and doesn't have to block this.

JeromeMartinez:
Sorry, I inadvertently deleted the branch.
Pinging about this PR, if not acceptable a is I would like to have hints about what would be acceptable.
My opinion is similar to the one from Dave (changing even more in order to be more readable would be part of another PR, this one is for carriage returns only).

michaelni:
this doesnt apply cleanly anymore

JeromeMartinez:
Branch rebased.

JeromeMartinez:
Branch rebased.

PR #78, opened 2017-07-02
escape underscore

dericed:


dericed:
This fixes the underscore according to the note in https://github.com/FFmpeg/FFV1/pull/68#issuecomment-312508276. @michaelni, I only found one occurrence of this to fix.
Note that in the LaTeX rendering of `$$$ x_y a\_b $$$` the `y` would be a subscript of `x` whereas `a_b` will have a literal underscore.

PR #82, opened 2018-01-02
Minor edits

dericed:


ffv1.md
```diff
@@ -160,13 +160,13 @@ Several components of FFV1 are described in this document using pseudo-code. Not
 
 ### NumBytes
 
-NumBytes is a non-negative integer that expresses the size in 8-bit octets of particular FFV1 components such as the Configuration Record and Frame. FFV1 relies on its container to store the NumBytes values, see [the section on the `Mapping FFV1 into Containers`](#mapping-ffv1-into-containers).
+`NumBytes` is a non-negative integer that expresses the size in 8-bit octets of particular FFV1 components such as the `Configuration Record` and Frame. FFV1 relies on its container to store the `NumBytes` values, see [the section on the `Mapping FFV1 into Containers`](#mapping-ffv1-into-containers).
```
retokromer: I would say either \`Frame\` or frame.

ffv1.md
```diff
@@ -593,7 +593,7 @@ The alternative state transition table has been built using iterative minimizati
 
 ### Huffman coding mode
 
-This coding mode uses Golomb Rice codes. The VLC code is split into 2 parts, the prefix stores the most significant bits, the suffix stores the k least significant bits or stores the whole number in the ESC case. The end of the bitstream (of the frame) is filled with 0-bits until that the bitstream contains a multiple of 8 bits.
+This coding mode uses Golomb Rice codes. The VLC code is split into 2 parts, the prefix stores the most significant bits, the suffix stores the k least significant bits or stores the whole number in the ESC case. The end of the bitstream of the frame is filled with 0-bits until that the bitstream contains a multiple of 8 bits.
```
retokromer: "0-bits" and "until that" sounds bizarre to me (not English native): possibly "filled with 0s until the bitstream contains"?

ffv1.md
JeromeMartinez: `Frame` is defined:
https://github.com/FFmpeg/FFV1/blob/master/ffv1.md#frame

so +1 for `Frame`

dericed:
I suggest to address the comments in a subsequent PR since they are not about the changes of the PR.

JeromeMartinez:
IMO the first one is related, as you put quotes for all items having a dedicated chapter for bitstream description (e.g. `Configuration Record`) except `Frame`. `Configuration Record` and `Frame` are linked (both are at the top level in a container)

dericed:
OK `Frame` is updated in https://github.com/FFmpeg/FFV1/pull/82/commits/5976f83882749d8dae2509891cf81a5db2957f02. There are a few uses of the term "frame" that don't refer to the FFV1 context of the term, so some of backtick quoted and some not.

dericed:
updated to avoid plural of `Frame`

michaelni:
Please use commit messages that do not need to be edited
4f6df662631ec328f41071a5174f8535b19534be for example says "plz review", i cannot push this as is and thus cannot just merge this without changes

-The bitstream contains 1 or more Quantization Table Sets.  
+The FFV1 bitstream contains 1 or more Quantization Table Sets.  
This adds trailing whitespace, i suggest you configure your editor to never add trailing whitespace

dericed:
@michaelni, the `plz review` commit message is removed.

The patch preserves the trailing whitespace before and after the edit. In markdown when a line ends with two-spaces (as it does here and a few other places), then a line break is inserted. Personally I'd prefer to use an actual line break rather than the two-space line endings. This is the topic of another PR at https://github.com/FFmpeg/FFV1/pull/77.

michaelni:
Merged, 
i had to rebase it as it wasnt based on latest master HEAD

thanks

PR #84, opened 2018-01-03
use version variable in makefile

dericed:


dericed:
slightly easier to maintain

PR #85, opened 2018-01-03
line breaks and formula formatting

dericed:


dericed:
I rebased this pr with the updated master branch. Please re-review.

michaelni:
Iam gettimng these warnings: with this locally merged
WARNING: artwork outdented 3 characters to avoid overrunning right margin around input line 1381)
WARNING: Output line (from source around line 1381) is 82 characters; longer than 72.  Excess characters: ' ] ) & 255':
  'initial_state[ i ][ j ][ k ] = ( pred + initial_state_delta[ i ][ j ][ k ] ) & 255'

dericed:
@michaelni I updated to add a line break in the middle of that equation. We already use this practice in a few other places such as the equations in the 'State transition table' section.

dericed:
bump :)

PR #90, opened 2018-01-04
Use Golomb Rice wording when applicable

JeromeMartinez:
Golomb Rice is used more often than VLC or Huffman coding, and are synonym here.


michaelni:
Is the term Golomb Rice actually correct ? iam not good at remembering these name - code relations but isnt this a exponential golomb code ? In the most strict interpretation huffman may actually be wrong too

michaelni:
> Is the term Golomb Rice actually correct ? 

And i succeeded in messing the terms up i think :) golomb rice is correct more or less

PR #91, opened 2018-01-04
Add get_bits() definition

JeromeMartinez:
There was no definition of this function.

Issue reported by @dwbuiten

### DIFF
--- a/ffv1.md
+++ b/ffv1.md
@@ -166,7 +166,7 @@ Several components of FFV1 are described in this document using pseudo-code. Not
 
 #### remaining_bits_in_bitstream
 
-`remaining_bits_in_bitstream( )` means the count of remaining bits after the current position in that bitstream component. It is computed from the `NumBytes` value multiplied by 8 minus the count of bits of that component already read by the bitstream parser.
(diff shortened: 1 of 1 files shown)


ffv1.md
```diff
@@ -172,6 +172,10 @@ Several components of FFV1 are described in this document using pseudo-code. Not
 
 `byte_aligned( )` is true if `remaining_bits_in_bitstream( NumBytes )` is a multiple of 8, otherwise false.
 
+#### get_bits
+
+`get_bits( i )` is the action to read the next `i` bits in the bitstream and to return the corresponding value. The current position is increased by `i`.
```
dericed: nit: better term for `current position`?

retokromer: I guess, I would use the term `pointer`.

JeromeMartinez: I reused `current position` from line 169.
Should I change the term in the same PR?

JeromeMartinez:
added a commit for changing "current position" to "pointer"

ffv1.md
dericed: I preferred the term `current position` over `pointer` but ¯\\_(ツ)_/¯

JeromeMartinez: blank vote, need another voter.

michaelni:
This definition seems unclear in relation to lest significant bit vs most significant bit in a byte first.

JeromeMartinez:
PR updated with addition of `from most significant bit to least significant bit`

PR #93, opened 2018-01-04
Removing v4 (draft) related stuff.

JeromeMartinez:
Should be put back in the a v4 branch, with more definitions (e.g. "PCM").

Issue reported by @dwbuiten

### DIFF
--- a/ffv1.md
+++ b/ffv1.md
@@ -811,10 +811,6 @@ SliceHeader( ) {                                              |
     picture_structure                                         | ur
     sar_num                                                   | ur
     sar_den                                                   | ur
-    if (version >= 4) {                                       |
-        reset_contexts                                        | br
-        slice_coding_mode                                     | ur
-    }                                                         |
 }                                                             |
 ```
 
@@ -871,22 +867,6 @@ MUST be 0 if sample aspect ratio is unknown.
 `sar_den` specifies the sample aspect ratio numerator.
 Inferred to be 0 if not present.
 MUST be 0 if sample aspect ratio is unknown.
-
-### reset_contexts
-
-`reset_contexts` indicates if slice contexts must be reset.
-Inferred to be 0 if not present.
-
-### slice_coding_mode
-
-`slice_coding_mode` indicates the slice coding mode.
-Inferred to be 0 if not present.
-
-|value  | slice coding mode            |
-|-------|:-----------------------------|
-| 0     | normal Range Coding or VLC   |
-| 1     | raw PCM                      |
-| Other | reserved for future use      |
 
 ## Slice Content
 
@@ -1268,7 +1248,7 @@ Note: 101376 is the frame size in pixels of a 352x288 frame also known as CIF ("
 
 For each `Frame`, each position in the slice raster MUST be filled by one and only one slice of the `Frame` (no missing slice position, no slice overlapping).
 
-For each `Frame` with keyframe value of 0, each slice MUST have the same value of slice\_x, slice\_y, slice\_width, slice\_height as a slice in the previous `Frame`, except if reset\_contexts is 1.
+For each `Frame` with keyframe value of 0, each slice MUST have the same value of slice\_x, slice\_y, slice\_width, slice\_height as a slice in the previous `Frame`.
 
 # Security Considerations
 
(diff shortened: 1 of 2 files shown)


dericed:
We should ensure this data doesn't get lost. I think better to wait a bit before forking this doc to start on a v4-specific specification, but perhaps for the meanwhile these removals should be accompanied by moving that info to a v4_notes.md document or into the issue tracker.

JeromeMartinez:
v4_notes.md added, may need a rebase after https://github.com/FFmpeg/FFV1/pull/95.

v4_notes.md
```diff
@@ -0,0 +1,4 @@
+# Notes for Version 4
+
+`slice_coding_mode` was defined in a former version of the spec, it could be peek for a former version (<201-01-04)
```
dericed: 2018?
Also I suggest to link to the commit with the removal in the note, i.e. https://github.com/FFmpeg/FFV1/pull/93/commits/046648bbbf6c8ad3f277b49d26d18e5038f527ec.

JeromeMartinez: Fixed.

ffv1.md
```diff
@@ -803,10 +803,6 @@ SliceHeader( ) {                                              |
     picture_structure                                         | ur
     sar_num                                                   | ur
     sar_den                                                   | ur
-    if (version >= 4) {                                       |
-        reset_contexts                                        | br
```
dericed: If `reset_contexts` is removed here, there remain a few stranded references to it.

JeromeMartinez: Fixed

JeromeMartinez:
PR updated with `reset_contexts` removed (and commit hash in v4_notes.md updated).
Note that nothing says that `reset_contexts` and `slice_coding_mode` would be in v4 (they are only a suggestion in FFmpeg for the moment), so v4_notes.md is only a reminder in case we need the old content.

dericed:
This patch also leaves a stray reference to reset_contexts in the restrictions section.

JeromeMartinez:
> This patch also leaves a stray reference to reset_contexts in the restrictions section.

Oops, fixed.

JeromeMartinez:
Rebased.

v4_notes.md
```diff
@@ -0,0 +1,4 @@
+# Notes for Version 4
+
+`reset_contexts` and `slice_coding_mode` were defined in a former version of the spec, it could be peek for a former version (<2018-01-04), commit d712fd1.
```
retokromer: You could give the link to the commit.

JeromeMartinez: chicken and egg: the reference Git repo is https://git.videolan.org/?p=ffmpeg.git and it does not have the commit yet.
Same for https://github.com/FFmpeg/FFV1 (a mirror) master branch, the only link available is in the PR https://github.com/FFmpeg/FFV1/pull/93/commits/0fe8b02af39a681e8862b279b2269bae0fc397b6 and I am reluctant to link to a GitHub PR commit.

retokromer: Ah, thanks, understood! I was thinking that a live link is more useful…

michaelni:
Isnt there some syntax with which these could be removed at "make" time ? so it all would stay in one place. That would ensure its not lost

dericed:
@michaelni that's may be a better idea for maintenance. We could reuse the process we have for math (`PDF:` and `RFC:` prefixes) and create `V013:` and `V4` prefixes. Any +1's?

dericed:
@retokromer imho a discussion on having V1 and V3 separated would have been better at the time the cellar charter was reviewed and approved. I suggest to follow https://datatracker.ietf.org/doc/charter-ietf-cellar/ and not separate V1 and V3.

dericed:
@JeromeMartinez are you in favoring of closing this in consideration of https://github.com/FFmpeg/FFV1/pull/102?

JeromeMartinez:
superseded by https://github.com/FFmpeg/FFV1/pull/102

retokromer:
> imho a discussion on having V1 and V3 separated would have been better at the time the cellar charter was reviewed and approved

My proposal was meant only for practical reasons. I don’t need it myself anymore, as I already do have fast encoders for both v1 and v3, but I would have been grateful for having it when I programmed that stuff. And, to avoid any misunderstanding, I am very fine with not separating them.

PR #96, opened 2018-01-04
Makes bits_per_raw_sample more generic

JeromeMartinez:
bits_per_raw_sample is also for Alpha plane and may be for new planes in the future.


ablwr:
+1 for broadening this scope, thanks!

PR #92, opened 2018-01-04
Clarify the difference between Pixel and Sample

JeromeMartinez:
There are sometimes inversion between the meaning of each word. Trying to clarify, with definition.

Issue reported by @dwbuiten

PR #97, opened 2018-01-04
change definitions from use of luminance to use of luma

ablwr:
As raised in #92, uses `luma` in definitions instead of `luminance`


JeromeMartinez:
I thought it would be more complicated.
Wondering it there are other places with similar issue about luma/chroma wording, ping @retokromer .

ablwr:
Everywhere else uses luma!

dericed:
nudging for either more reviews or a merge on this and the others. I'd like to get an active ffv1 draft back in the cellar tracker.

PR #98, opened 2018-01-06
bump to version 01

dericed:


dericed:
I suggest to stall merging this PR until after finalizing and merging the other open pulls (perhaps excepting #71 which needs a lot more work). The cellar document tracker's version of the FFV1 specification is now expired https://datatracker.ietf.org/doc/draft-ietf-cellar-ffv1/ so we don't have an active spec listed, but I think the recent work is worthy of marking a new version. After merging please also mark a release in the github repo.

dericed:
I propose resolving https://github.com/FFmpeg/FFV1/pull/89 and maybe https://github.com/FFmpeg/FFV1/pull/87 and then merging this, so that the tracker can be updated. Following that we can adjust the syntax and makefile to separate a [v013] and [v4] release from the same document.

JeromeMartinez:
https://github.com/FFmpeg/FFV1/pull/101 is not big and is the missing link between pseudo-algorithm and Samples description, I suggest to have it merged too before marking a new version.
Eager to send this new version on IETF tracker.

dericed:
I suggest that other open PR's either relate to future versions or require more discussion. Thus I propose merging this PR so that the cellar document tracker may be updated with version 01.

JeromeMartinez:
From my point of view https://github.com/FFmpeg/FFV1/pull/87 and https://github.com/FFmpeg/FFV1/pull/93 are waiting for feedback or merge but I think this is not blocking for bump to version 01 on the IETF tracker.

dericed:
I prefer an alternate to https://github.com/FFmpeg/FFV1/pull/93 to mark lines that are `[V4]`-only or `[V013]`-only and handle filtering lines in the make process. 

https://github.com/FFmpeg/FFV1/pull/87 looks ok to me, but perhaps needs more consensus on the terminology used. At any rate I'd prefer to move forward with this PR and not consider the other PRs as necessarily blocking.

dericed:
thx @michaelni, please also mark a release point in the github repo for `draft-ietf-cellar-ffv1-01`

michaelni:
done: https://github.com/FFmpeg/FFV1/releases/tag/draft-ietf-cellar-01

PR #102, opened 2018-02-03
support method to output lines only to version0,1,3 or version4

dericed:
I consider this strategy temporary until version 0,1,3 is finalized by
IETF.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,29 +2,48 @@ SRC=ffv1.md
 PDF=$(SRC:.md=.pdf)
 HTML=$(SRC:.md=.html)
 VERSION=01
+VERSION-v4=00
 
 $(info PDF and HTML rendering has been tested with pandoc version 1.13.2.1, some older versions are known to produce very poor output, please ensure your pandoc is recent enough.)
 $(info RFC rendering has been tested with mmark version 1.3.4 and xml2rfc 2.5.1, please ensure these are installed and recent enough.)
 
-all: ffv1.html ffv1.pdf draft-ietf-cellar-ffv1-$(VERSION).html draft-ietf-cellar-ffv1-$(VERSION).txt
-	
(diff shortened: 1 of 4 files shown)


dericed:
This PR is intended to supersede https://github.com/FFmpeg/FFV1/pull/93.

michaelni:
breaks build:

pandoc --toc -s --number-sections --pdf-engine=pdflatex -V geometry:margin=1in --variable urlcolor=blue -o "ffv1.pdf" merged_ffv1pdf.md
pandoc: unrecognized option `--pdf-engine=pdflatex'
Try pandoc --help for more information.
make: *** [ffv1.pdf] Error 2

michaelni:
pandoc --version
pandoc 1.19
Compiled with pandoc-types 1.17.0.4, texmath 0.9, highlighting-kate 0.6.3
Default user data directory: /home/michael/.pandoc
Copyright (C) 2006-2016 John MacFarlane

dericed:
Hi @michaelni, for me I had to change `--latex-engine=xelatex` to `--pdf-engine=pdflatex` else I would get
```
--latex-engine has been removed.  Use --pdf-engine instead.
Try pandoc --help for more information.
```

but I'm on pandoc 2.1.1. I changed it back so as not to conflate the issues in a single PR, but I suggest we consider updating the pandoc dependency to a more modern version.

The PR is updated for pandoc 1.19 for a re-review.

dericed:
I added one more commit to split the title and use "FFV1 Lossless Video Format Version 0, 1, and 3" and "FFV1 Lossless Video Format Version 4" rather than "FF Video Codec 1". IMHO "FF Video Codec 1" is an unfamiliar term and using FFV1 in the title would be better.

JeromeMartinez:
I suggest  "FFV1 Video Coding Format Version (...)", like in https://en.wikipedia.org/wiki/AV1

dericed:
@JeromeMartinez i agree, updated

PR #108, opened 2018-02-12
fix inter-sectional linking

dericed:


dericed:
addresses comment in https://github.com/FFmpeg/FFV1/pull/71#issuecomment-365087119

dericed:
It seems in mmark, that underscores are simply removed in order to support linking to other sections with underscores in the section name.

PR #112, opened 2018-04-21
Version 0 and 1 can have some content after SliceContent

JeromeMartinez:
From the debate in https://github.com/FFmpeg/FFV1/pull/110, I propose a different method for avoiding to have streams created by ` "ffmpeg -c ffv1 -level 1 -slicecrc 1` considered as non conforming to specification.

This patch synchronizes specification with behavior of the decoder (which ignores the extra bits) by permitting the extra bits. It also add a note for people who would revise the specification about the risk of using 40 bits of the reserved part.

This patch does not prevent us to change our mind about https://github.com/FFmpeg/FFV1/pull/110 in the future.

Additional note: the extra bits are authorized only for versions 0 and 1 due to the existence of such bistream (not expected, but they are there) in the wild. "feature" of not having extra bits in version 3 is used by some decoders for being able to decode first slices (when latest bytes of a frame are missing, FFmpeg decoder decodes only the first slice, my own decoder uses the lack of reserved bits in the bitstream for parsing all slices up to the missing bytes by starting from the beginning and parsing all slices sequentially).

### DIFF
--- a/ffv1.md
+++ b/ffv1.md
@@ -782,6 +782,10 @@ Slice( ) {                                                    |
     if (coder_type == 0)                                      |
         while (!byte_aligned())                               |
             padding                                           | u(1)
+    if (version <= 1) {                                       |  
+        while (remaining_bits_in_bitstream( NumBytes ) != 0 ) |
+            reserved                                          | u(1)
+    }                                                         |
     if (version >= 3)                                         |
         SliceFooter( )                                        |
 }                                                             |
@@ -789,6 +793,11 @@ Slice( ) {                                                    |
 
 `padding` specifies a bit without any significance and used only for byte alignment.
 MUST be 0.
+
+`reserved` specifies a bit without any significance in this revision of the specification and may have a significance in a later revision of this specification.  
+Encoders SHOULD NOT fill these bits.  
+Decoders SHOULD ignore these bits.  
+Note in case these bits are used in a later revision of this specification: any revision of this specification SHOULD care about avoiding to add 40 bits of content after `SliceContent` for version 0 and 1 of the bitstream. Background: due to some non conforming encoders, some bitstreams where found with 40 extra bits corresponding to `error_status` and `slice_crc_parity`, a decoder conforming to the revised specification could not do the difference between a revised bitstream and a buggy bitstream.
 
 ## Slice Header
 


JeromeMartinez:
the patch was made before https://github.com/FFmpeg/FFV1/pull/110#issuecomment-383283621 comment, let me know if "SHOULD" should be replaced by "MUST". IMO "SHOULD" is better here as there may be a good reason e.g. for checking CRC with "./ffmpeg -f lavfi -i mandelbrot -t 1 -c ffv1 -slicecrc 1 a.mkv" (current FFmpeg), but I am not strongly against using "MUST", priority is that files created this way are not considered invalid by the spec.

IETF meaning of "SHOULD": "there may exist valid reasons in particular circumstances to ignore a particular item"

michaelni:
I dont think extensibility and sequential decodability should be or are exclusive.
the "shape and amount of fields" written would depend on the version. so any decoder that is implemented after this knows what the fields are and does not see any reserved fields, thus it can just read sequentially. A decoder thats older would see the same data as reserved fields, such a decoder is not really up to date and it would have to find and decode the slices by following the pointers. Does this resolve your concern about limiting / not limiting this to specific versions ? or am i missing something ?

JeromeMartinez:
You speak of version 3+, right?
Currently, there are no "reserved fields", so a decoder can not catch "reserved fields".

>  such a decoder is not really up to date and it would have to find and decode the slices by following the pointers.

This is my concern: if we add new fields, we break the idea of being backward compatible with older decoders which fallback to sequential decoding when a frame is corrupted, and IMO this is not good. I remember I asked some time ago about the capability to decode sequentially, and the current spec permits that. We should not break decoders, even if they are old and handle "new" v3 streams, else we'll break the trust in the format.

Permitting new fields after data should be planned in v4, if necessary, by showing how new fields are introduces without breaking compatibility. Else we need to explicitly indicate that a new micro_version has an impact on sequential decoding (I will be against this change for v3, as v3 sequential decoders are already in the wild and micro_version is not tested for knowing if sequential decoding is possible, as it was not indicated in the draft; this is a draft, true, but this item was already discussed, so I don't see a good reason to break the idea).

Anyway, this PR is for v0/v1, and here this is different (no slices, and frame byte size is provided by external method), so doable without risks. I suggest to merge the PR for v0/v1, then we debate separately about v3 (this PR does not touch on v3 so it introduces no change on it)

Issue #115, opened 2018-06-01
please mark a release for draft-ietf-cellar-ffv1-03


dericed:

michaelni:
done, https://github.com/FFmpeg/FFV1/releases/tag/draft-ietf-cellar-ffv1-03

PR #116, opened 2018-06-01
Review ietf cellar ffv1 02 secdir early xia 2018 06 01

dericed:
addresses many of the comments in https://datatracker.ietf.org/doc/review-ietf-cellar-ffv1-02-secdir-early-xia-2018-06-01/.

https://github.com/FFmpeg/FFV1/issues/115 should be resolved before this is merged.


ffv1.md
```diff
@@ -476,7 +474,7 @@ RFC:```
 RFC:j_{0} = 2
 RFC:```
 
-#### Range non binary values
+#### Range Non Binary Values
```
retokromer: I am not an English speaker: `Non Binary` or `Non-Binary`?

dericed: IIRC hyphens presents in section headers break the inter-sectional linking as the document is processed through mmark and xml2rfc.

retokromer: Thank you for the explanation, @dericed!

PR #121, opened 2018-07-26
bump version to ffv1-04, ffv1-v4-01

dericed:
Given @mcr's comments "Internet Drafts are cheap documents, please do not treat them as more than
"draft"s... submit often.", I propose a new release. I can them update the cellar docs and follow up on the review which led to most of the recent commits. I suggest following this merge with a new release `draft-ietf-cellar-ffv1-02_v4-01`.


kieranjol:
I've no comments aside from an appreciation of the @mcr quote.

PR #124, opened 2018-09-24
More details about the different parts of the stream

JeromeMartinez:
During review by IETF people, it was requested to add some intro for each paragraph.

### DIFF
--- a/ffv1.md
+++ b/ffv1.md
@@ -26,11 +26,8 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 ## Definitions
 
 -------- --------------------------------------------------------------
-`Frame`: An encoded representation of a complete static image.
 
-`Slice`: A spatial sub-section of a `Frame` that is encoded separately from an other region of the same frame.
-
-`Container`: Format that encapsulates `Frames` and (when required) a `Configuration Record` into a bitstream.
+`Container`: Format that encapsulates `Frames` (see [the section on `Frames`](#frame)) and (when required) a `Configuration Record` into a bitstream.
 
 `Sample`: The smallest addressable representation of a color component or a luma component in a `Frame`. Examples of sample are Luma, Blue Chrominance, Red Chrominance, Alpha, Red, Green, and Blue.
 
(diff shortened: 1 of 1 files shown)


ffv1.md
```diff
@@ -1210,7 +1227,9 @@ RFC:Its value is `floor(slice_x * frame_pixel_width / num_h_slices)`.
 
 ## Slice Footer
 
-Note: slice footer is always byte aligned.
+A `Slice Footer` provides information about slice size and (optionnally) parity.
```
dericed: optionally

ffv1.md
```diff
@@ -1009,6 +1014,10 @@ Architecture overview of slices in a `Frame`:
 
 ## Slice
 
+A `Slice` is an independant part of a frame. `Slice` are useful for multithread decoding.
```
kieranjol: Independent
Also perhaps 'mulithreaded'?

JeromeMartinez:
Typos, sorry :(.
Thanks, fixed.

ffv1.md
```diff
@@ -1009,6 +1014,10 @@ Architecture overview of slices in a `Frame`:
 
 ## Slice
 
+A `Slice` is an independent part of a frame. `Slice` are useful for multithreaded decoding.
```
dericed: Grammar here is a bit off. Also I think the definition of Slice within the Definition section at the beginning is better. Rather than define it twice perhaps the one in the Definitions section could be moved here. For the second sentence, maybe better to say like "The use of multiple `Slices` per frame can be useful for taking advantage of the opportunities of multithreaded encoding and decoding."

ffv1.md
```diff
@@ -975,6 +978,8 @@ FFV1 SHOULD use `V_FFV1` as the Matroska `Codec ID`. For FFV1 versions 2 or less
 
 ## Frame
 
+A `Frame` is the representation of video signals in the spatial domain. The whole `Frame` is provided by the underlaying container.
```
dericed: same comment as below, that perhaps the definition of Frame from the Definitions section should be moved here, so as not to twice Frame twice using separate language.

ffv1.md
```diff
@@ -703,6 +703,9 @@ Default values at the decoder initialization phase:
 
 ## Parameters
 
+A `Parameters` element provides information about the decoding configuration of the frame (version 0 and 1) or the whole stream (other versions).
```
dericed: backtick quote Frame. Also throughout the document it seems we've been preferring to say 'bitstream' (or 'bytestream') rather than simply 'stream'.

dericed:
I updated this commit according to the comments. I removed the initial definitions of Slice and Frame since now they are defined more extensively in their own sections. Because of this change the first references to `Slice` and `Frame` include forward references. Ready for re-review.

PR #127, opened 2018-09-25
Add an exception for sar_den when aspect ratio is unknown

### LINKED ISSUE #109: sar_den is 0 if unknown
In FFmpeg sar_num/sar_den for unknown sar is `0/1` but the [spec at this section](https://github.com/FFmpeg/FFV1/blob/aadbb8c486a6cf688d18ee18a7e329ec527a8268/ffv1.md#sar_den) says "MUST be 0". As I read, it:

`sar_num  = 0 &&  sar_den  = 0` valid
`sar_num != 0 && sar_den != 0` valid
`sar_num  = 0 && sar_den != 0` not valid
`sar_num != 0 &&  sar_den  = 0` not valid

Which is wrong?
[ ] FFmpeg
[ ] FFV1 specification
[ ] my interpretation

I suggest requiring a denominator to be 0 is not ideal and that we be more flexible. Unknown sar could be communication by the numerator alone.

JeromeMartinez:
Fix #109.
I prefer to explicitly handle the value used by FFmpeg instead of [the other proposal](https://github.com/FFmpeg/FFV1/issues/109#issuecomment-370037367), but I would not be against the other proposal if preferred by others.

### DIFF
--- a/ffv1.md
+++ b/ffv1.md
@@ -1106,13 +1106,17 @@ Inferred to be 0 if not present.
 
 `sar_num` specifies the sample aspect ratio numerator.  
 Inferred to be 0 if not present.  
-MUST be 0 if sample aspect ratio is unknown.
+A value of 0 means that aspect ratio is unknown.  
+Encoders MUST write 0 if sample aspect ratio is unknown.  
+If `sar_den` is 0, decoders SHOULD ignore the encoded value and consider that `sar_num` is 0.
 
 ### sar_den
 
 `sar_den` specifies the sample aspect ratio denominator.  
 Inferred to be 0 if not present.  
-MUST be 0 if sample aspect ratio is unknown.
+A value of 0 means that aspect ratio is unknown.  
+Encoders MUST write 0 if sample aspect ratio is unknown.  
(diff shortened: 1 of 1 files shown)


dericed:
Would it make any difference to the decoder if the sar is 0/1 vs 0/0? Since in both case the num is 0, then the spec says the sar is unknown. I hesitate on this since telling a decoder that a denominator should be zero sounds like a problem.

JeromeMartinez:
I personally prefer to have a single pair 0/0 for unknown SAR, so considering any other value (except 0/1 due to the streams in the wild) as a potentially corrupted stream, but I am fine with considering 0/n as unknown SAR.
@michaelni your preference?

dericed:
+1 for 0/n meaning unknown.

michaelni:
0/n and n/0 are equivalent by symmetry. In fact i would not be surprised if broken software which wrote one wouldnt be writing the other too when a video is rotated by 90°.
IMHO, n/m n,m>0 is an aspect ratio, 0/0 is unknown, other values are undefined but should be interpreted as unknown if encountered

JeromeMartinez:
PR updated by handling 0/n and n/0 the same way.

PR #129, opened 2018-10-02
add link to version 4

retokromer:


retokromer:
Conflict resolved.

retokromer:
You can rebase when merging. (Otherwise I will do it as soon as I will be again on a «normal» computer.)

PR #130, opened 2018-10-03
More minor edits

dericed:
a mix of commits to add definition, make referenced concepts more consistent, and some rewording for clarity.

### DIFF
--- a/ffv1.md
+++ b/ffv1.md
@@ -190,17 +190,17 @@ a = b, a += b, a -= b, a *= b
 
 # Sample Coding
 
-Samples within a plane are coded in raster scan order (left->right, top->bottom). Each sample is predicted by the median predictor from samples in the same plane and the difference is stored see [Coding of the Sample Difference](#coding-of-the-sample-difference).
+For each `Slice` (as described in [the section on `Slices`](#slice)) of a `Frame`, the planes, lines, and samples are coded in an order determined by the `Color Space` (see [the section on `Color Space`](#color-space)). Each Sample is predicted by the median predictor as described in [the section of the Median Predictor](#median-predictor) from other samples within the same plane and the difference is stored using the method described in [Coding of the Sample Difference](#coding-of-the-sample-difference).
 
 ## Border
 
-A border is assumed for each coded `Slice` (see [the section on `Slices`](#slice)) for the purpose of the median predictor and context according to the following rules:
+A border is assumed for each coded `Slice` for the purpose of the median predictor and context according to the following rules:
 
 - one column of samples to the left of the coded slice is assumed as identical to the samples of the leftmost column of the coded slice shifted down by one row. The value of the topmost sample of the column of samples to the left of the coded slice is assumed to be `0`
 - one column of samples to the right of the coded slice is assumed as identical to the samples of the rightmost column of the coded slice
 - an additional column of samples to the left of the coded slice and two rows of samples above the coded slice are assumed to be `0`
 
-The following table depicts a slice of samples `a,b,c,d,e,f,g,h,i` along with its assumed border.
+The following table depicts a slice of 9 samples `a,b,c,d,e,f,g,h,i` in a 3x3 arrangement along with its assumed border.
 
 ```
 +---+---+---+---+---+---+---+---+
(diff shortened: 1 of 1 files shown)


ffv1.md
```diff
@@ -31,6 +31,8 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
 `Sample`: The smallest addressable representation of a color component or a luma component in a `Frame`. Examples of sample are Luma, Blue Chrominance, Red Chrominance, Alpha, Red, Green, and Blue.
 
+`Plane`: A discrete component of a static image comprised of samples that represent a specific quanitification of luma, chroma, or alpha visual information at that image.
```
JeromeMartinez: I would prefer a more generic description, planes could have something else in the future e.g. Green Difference with Bayer formats. `luma, chroma, or alpha visual information` could be replaced by `Sample` (this word is used in the spec at several places).
Typo `quanitification`.

ffv1.md
```diff
@@ -186,19 +188,19 @@ a = b, a += b, a -= b, a *= b
 
 `get_bits( i )` is the action to read the next `i` bits in the bitstream, from most significant bit to least significant bit, and to return the corresponding value. The pointer is increased by `i`.
 
-# General Description
+# Sample Encoding
 
-Samples within a plane are coded in raster scan order (left->right, top->bottom). Each sample is predicted by the median predictor from samples in the same plane and the difference is stored see [Coding of the Sample Difference](#coding-of-the-sample-difference).
+For each `Slice` (as described in [the section on `Slices`](#slice)) of a `Frame`, the samples are coded in an order determined by the `colorspace_type` (see [the section on `colorspace_type`](#colorspacetype)). Each sample is predicted by the median predictor as described in [the section of the Median Predictor](#median-predictor) from other samples within the same plane and the difference is stored using the method described in [Coding of the Sample Difference](#coding-of-the-sample-difference).
```
JeromeMartinez: scan order (left->right, top->bottom) is no more indicated (`colorspace_type` does not indicate such order, it is common to all `colorspace_type`), it should be indicated somewhere else if not here.

ffv1.md
```diff
@@ -678,6 +680,10 @@ Note, this is different from JPEG-LS, which doesn’t use prediction in run mode
 
 # Bitstream
 
+An FFV1 bitstream is composed of a series of 1 or more `Frames` and (when required) a `Configuration Record`.
```
JeromeMartinez: I would invert `Configuration Record` and `Frames` here, because `Configuration Record` must be present first and it IMO looks a bit weird not to indicate it first in this sentence.

ffv1.md
dericed: I was concerned that `scan order (left->right, top->bottom)` was misleading since the samples are ordered like:
- RGB is left->right, first plane->last plane, top->bottom
- YCbCr is first plane->last plane, left->right, top->bottom

JeromeMartinez: if we go deep in details, the order is:

- RGB is for each line (top->bottom), list planes (Y, U, V), and for each plan list samples (left->right)
- YCbCr is for each plane (Y, Cb, Cr), list lines (top->bottom), and for each line list samples (left->right)

I understand the need to remove it from here, but I don't find elsewhere that lines listing is top to bottom and sample listing is left to right, so IMO it is needed to indicate it somewhere else before removing the same information from this line.

dericed: updated, I moved a description of the order to the beginning of the section of Color Spaces

dericed:
just updated in order to squash some commits.

michaelni:
pushed the first commit of the series. About the 2nd, I think your definition of plane is confusing.

ffv1.md
```diff
@@ -186,19 +188,19 @@ a = b, a += b, a -= b, a *= b
 
 `get_bits( i )` is the action to read the next `i` bits in the bitstream, from most significant bit to least significant bit, and to return the corresponding value. The pointer is increased by `i`.
 
-# General Description
+# Sample Encoding
```
michaelni: The specification describes the format and how to decode it. The word "encoding" is a bit confusing as it suggests that how to encode would be specified. In fact one could design a lossy encoder that outputs the same format that a lossless decoder decodes

michaelni:
Also i think its not a good idea to replicate text alot. the RCT and the predictor could easily change in future versions theres more to rewrite/update in that case.

ffv1.md
dericed: I found the term "General Description" to be very vague for the contents. The subsections here all seem to contextualize how samples are coded. Would `Sample Coding` work?

michaelni: Sample Coding sounds ok to me

dericed: updated that commit to https://github.com/FFmpeg/FFV1/pull/130/commits/89e45281bc89e74dde860fb4ac7c0c0af402a346

PR #138, opened 2019-01-10
[WIP] Multispectral support

michaelni:
This is not to merge yet, but for discussion and of course if anyone wants to contribute/add to it

### DIFF
--- a/ffv1.md
+++ b/ffv1.md
@@ -731,10 +731,32 @@ Parameters( ) {                                               |
     colorspace_type                                           | ur
     if (version >= 1)                                         |
         bits_per_raw_sample                                   | ur
-    chroma_planes                                             | br
-    log2_h_chroma_subsample                                   | ur
-    log2_v_chroma_subsample                                   | ur
-    extra_plane                                               | br
+    if (colorspace_type < 2) {                                |
+        chroma_planes                                         | br
+        log2_h_chroma_subsample                               | ur
+        log2_v_chroma_subsample                               | ur
+        extra_plane                                           | br
+    } else {                                                  |
+        plane_count                                           | ur
+        per_plane_metadata_count                              | ur
+        for( i = 0; i < plane_count; i++ ) {                  |
+            plane_type                                        | ur
+            center_frequency                                  | ur
+            frequency_response ( center_frequency )           |
+            color_transfer_characteristic                     | ur
+            black_point                                       | sr
+            white_point                                       | sr
+            unit                                              | sr
+            direction_x                                       | sr
+            direction_y                                       | sr
+            direction_z                                       | sr
+            h_subsample                                       | ur
+            v_subsample                                       | ur
+            for (j = 0; j < per_plane_metadata_count; j++) {  |
+                reserved[j]                                   | sr
+            }                                                 |
+        }                                                     |
+    }                                                         |
     if (version >= 3) {                                       |
         num_h_slices - 1                                      | ur
         num_v_slices - 1                                      | ur
@@ -823,6 +845,7 @@ If state_transition_delta is not present in the FFV1 bitstream, all Range coder 
 |-------|:------------------------|:------------------------|:------------------------|:------------------------|
 | 0     | YCbCr                   | None                    | Transparency            | `Plane` then `Line`     |
 | 1     | RGB                     | JPEG2000-RCT            | Transparency            | `Line` then `Plane`     |
+| 2     | Multispectral           | Custom                  | Transparency            | TODO/WIP/XXX            |
 | Other | reserved for future use | reserved for future use | reserved for future use | reserved for future use |
 
 Restrictions:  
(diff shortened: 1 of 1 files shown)


ffv1.md
```diff
@@ -731,10 +731,32 @@ Parameters( ) {                                               |
...
+        log2_v_chroma_subsample                               | ur
+        extra_plane                                           | br
+    } else {                                                  |
+        plane_count                                           | ur
+        per_plane_metadata_count                              | ur
+        for( i = 0; i < plane_count; i++ ) {                  |
+            plane_type                                        | ur
+            center_frequency                                  | ur
+            frequency_response ( center_frequency )           |
+            color_transfer_characteristic                     | ur
```
JeromeMartinez: `color_transfer_characteristic ` is common to YUV (+colour_primaries, matrix_coefficients), shouldn't it be in a color info dedicated part e.g. at the end of the configuration box?

retokromer:
On my side, I made good experiences definitying the wavelength (in nm) for each spectral band.

ffv1.md
michaelni: hmm, the idea was that we need something to go from physical energy to binary values and that often is not linear. Maybe color_transfer_characteristic is a poor term for this.

JeromeMartinez: `transfer_characteristic` wording would be misleading as the name is already used elsewhere (prefix 'color' would not be enough for differentiation IMO), so in that case it should be another name.

michaelni:
> On my side, I made good experiences definitying the wavelength (in nm) for each spectral band.

I used hz in the hope that the whole relevant EM spectrum could be represented and as its a fundamental unit. (quite possibly ive missed something of course). with nm we would need fractional values and not just integers if we wanted to store X-ray images and use this field to hold some acccurate X-ray frequency information

retokromer:
I’m afraid, I’m probably missing something! It seems to me that the integers vs. fractions issue is the same when expressed in Hz or in m, which are actually both SI base units. (Context: I made that choice, because it’s the information characterising the filters we use… Anyway, the conversion between them is elementary.)

michaelni:
If one wanted to use integers to handle the value, with Hz that works for most of the relevant spectrum with also very good precision for the range one is able to make pictures with. visual, UV, ir, xrays are all high frequency so lots of digits precission. With really long radio waves making images will be very hard as the antenna has to scale up size wise. OTOH with meters as the base unit most of the relevant spectrum falls below 1.0 so fractional numbers need to be handled. Also with wavelength we would need to clarify if that is wavelength in air, in vacuum or in the medium the picture was taken.
But ultimately it of course makes no real difference either way can be used to store the value.
In practice maybe we should use a mantisse * 2^exponent style storage as fractions as in a/b would result in values for either a or b not fitting in 64bit integers. Ill try to come up with some better suggestion in a future patch revission

ffv1.md
```diff
@@ -868,6 +891,78 @@ RFC:`log2_v_chroma_subsample` indicates the subsample factor, stored in powers t
...
+| 10      | Polarization                                    |
+| 11-999  | Reserved                                        |
+
+### per_plane_metadata_count
+
+`per_plane_metadata_count` allows future extension of the information we store about planes
+
+### center_frequency
+
+`center_frequency` indicates the center frequency in hz. Or 0 if unknown or not applicable.
```
retokromer: nit `Hz`

retokromer:
Just out of curiosity: is there any work going on here? (I will be in Vienna from Thursday to Sunday and could meet if useful. Then again the week 6–10 January.)

ffv1.md
pjotrek-b: Sorry for jumping in (3 years+ later) :smile: :
Does any of you have any particular real-world examples in mind (or access to them)?
Like the xrays, long radio wave images, air/vacuum/etc recordings @michaelni mentioned?

I'm wondering if it'd make sense not to raw-parametrize this, but rather define/offer presets which can be tuned in the code, rather than as arguments? Then access these interpretation options by a given name shortcut?
(If I understood the subject alright)

I assume that any area of expertise that would create such recordings will very likely have some sets of "common parameters" - therefore domain-specific "insider terms" for those imaging types.

Like FFmpeg having "dvd_pal" containing a set of sane/valid parameters, we could have something like "xray_10nm" or "10nm-water-hot". Dunno. Something like that?


Wouldn't this also allow arbitrary choice of data-types (per imaging-type) for proper calculations?
Integers, fractions, etc - different strokes for different (imaging type) folks so to say :stuck_out_tongue_winking_eye: 


If I'm right, then this is tricky to handle in a Standards paper definition... But if x264 can be a standard and do presets, so could ffv1, I presume? :wink:

ffv1.md
pjotrek-b: * `color_translation`?
  * `color_conversion`?
  * `color_interpretation`?

possibly add `_characteristic` to any of those?

ffv1.md
michaelni: If you need samples there should be lots of public stuff from astronomy and satellites. For xrays and MRT the word to search for is probably dicom image. One source for public domain samples would be any images you own the rights to, like from a medical exam. I didnt really try to find free proper raw sample images for these things previously ...

kieranjol:
I was at a pronom workshop recently and there were  some Dicom images.
Perhaps pronom have these images available as part of a test suite?

retokromer:
FYI: I will present on [this topic](https://reto.ch/training/2022/20221028/) again in a larger context at [NTTW6](https://mediaarea.net/NoTimeToWait6).

PR #139, opened 2019-01-17
uniform pseudo-code

retokromer:
The majority of our pseudo-code do not terminate an instruction by a semi-colon.

### DIFF
--- a/ffv1.md
+++ b/ffv1.md
@@ -516,21 +516,21 @@ pseudo-code                                                   | type
 --------------------------------------------------------------|-----
 void put_symbol(RangeCoder *c, uint8_t *state, int v, int \   |
 is_signed) {                                                  |
-    int i;                                                    |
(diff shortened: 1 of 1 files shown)


JeromeMartinez:
This part is the C source code, if we remove the C stuff, there are more to remove e.g. all the "int" declarations.

retokromer:
Currently in the specification these parts are declared as being pseudo-code, not as C code. Do you prefer if I delete the declarations as well or should I modify the headings instead?

In the past, I expressed my preference of using plain C, but the majority was against that suggestion. In my personal opinion, we should use constantly one through the full document.

JeromeMartinez:
> Do you prefer if I delete the declarations 

what I had in mind in case the ";" are removed, right.

retokromer:
Superseded by https://github.com/FFmpeg/FFV1/pull/140

PR #140, opened 2019-01-17
fix table title

retokromer:
Usually we have pseudo-code, but two times we use actual C-code excerpts.


dericed:
This changes makes it seem as if https://github.com/FFmpeg/FFV1/blob/master/ffv1.md#pseudo-code no longer applies to this sections, since the pseudo-code label is removed. Is that intended?

retokromer:
For consistency, if pseudo-code is used, then it should be used always. I suggest to not mix two philosophy in this short document.

dericed:
Before: pseudo code declaration applies to all labelled pseudo-code, but some pseudo-code is C
After: pseudo code declaration applies to all labelled pseudo-code, but there are some uncontextualized sections

i think here the pr changes the problem rather than resolves it. nudge to @michaelni

retokromer:
OK

dericed:
I think this is a case where the pr was closed before the pr was finished, I can move some pieces of this to an issue.

retokromer:
Possibly https://github.com/FFmpeg/FFV1/pull/139 is more useful.

retokromer:
Which one???

PR #145, opened 2019-02-04
bump version to ffv1-07_v4-04

dericed:
Since we're now in last call stage, let's get our published document up to date.


dericed:
When merged, please mark a release point as well.

PR #157, opened 2019-08-26
fix name of author

retokromer:


JeromeMartinez:
> This file is borrowed from elsewhere, so a change would be better upstream.

Looks like a UTF-8 issue (UTF-8 file handled as Latin1 at some point in the process) of the file in this repo, I bet that upstream has not the issue.

> Also the repo no longer uses this file,

File should be deleted?

JeromeMartinez:
> > Also the repo no longer uses this file,
> File should be deleted?

Sorry, it was already deleted in [this commit](https://github.com/FFmpeg/FFV1/commit/d188994cb84900cdd0c83c06f75656753217a911) merge not a long time ago, comment no more valid.

retokromer:
Upstream it’s correct.

PR #158, opened 2019-08-31
Nits and review responses

dericed:
These changes help draft-ietf-cellar-ffv1 pass https://tools.ietf.org/tools/idnits/ with no errors. The changes also take into account a discuss on tools-discuss on appropriateness of non-ascii characters in various contexts.

### DIFF
--- a/ffv1.md
+++ b/ffv1.md
@@ -1055,16 +1055,16 @@ Inferred to be 0 if not present.
 In the case of a FFV1 bitstream with `version >= 3`, a `Configuration Record` is stored in the underlying `Container`, at the track header level. It contains the `Parameters` used for all instances of `Frame`. The size of the `Configuration Record`, `NumBytes`, is supplied by the underlying `Container`.
 
 ```c
-pseudo-code                                                   | type
---------------------------------------------------------------|-----
-ConfigurationRecord( NumBytes ) {                             |
-    ConfigurationRecordIsPresent = 1                          |
-    Parameters( )                                             |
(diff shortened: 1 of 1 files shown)


retokromer:
floor(x) = ⌊x⌋ 
ceil(x) = ⌈x⌉

richardpl:
Yes, the previous explanation was reversed, this patch actually corrected that error, I got confused with this.

retokromer:
>Only the first two looks wrong, where are they explained.

Yep, @richardpl, you are right. I revoke my approval.

dericed:
@richardpl, I added a note about this on the corresponding commit, see https://github.com/FFmpeg/FFV1/pull/158/commits/8e1ef6d99ab5adbb0006951d4ffce0ef422a728b.

dericed:
ping to @mcr as this builds upon the discussion in tools-discuss

michaelni:
The definition for ceil/floor looks wrong after this.

Also somehow i did like the unicode symbols (when they where displayed correctly) but that cant be helped if thats not allowed ...

michaelni:
merged the first commit in the series, so theres 1 less now

dericed:
Just fixed the issue in https://github.com/FFmpeg/FFV1/pull/158#issuecomment-527515815 and force pushed.

michaelni:
is the commit message for the floor/ceil commit still accurate , or does that need a change too ?

dericed:
@michaelni, commit msg is now fixed, thanks for catching that.

dericed:
:) thx

PR #159, opened 2019-08-31
adds a step to encode and embed svg image data into the rfc xml

dericed:
This allows the RFC xml to be independent without needing the sidecar svg files and thus can be used in IETF rfc2xml version 3 tools such as https://xml2rfc.tools.ietf.org/experimental.html

This PR uses the 3rd option described in https://tools.ietf.org/html/rfc7991#section-2.5 for embedding svg image data into the artwork tag. The svg xml encoded is done by piping it through `python -c "import urllib;print urllib.quote(raw_input())"`. I wasn't able to find a way to do this within the Makefile: search the xml for svg file references, encode the svg xml, and then to update the xml file from the image file reference to the image encoded data. So I opted to do this in a sidecar bash file. Suggestions welcome, though this works for me.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ HTML=$(SRC:.md=.html)
 VERSION=08
 VERSION-v4=05
 
-$(info RFC rendering has been tested with mmark version 2.0.48, xml2rfc 2.23.1, xmlstarlet 1.6.1, pdfcrop v1.38, and pdf2svg 0.2.3, please ensure these are installed and recent enough.)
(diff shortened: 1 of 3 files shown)


dericed:
If https://github.com/FFmpeg/FFV1/pull/160 can be considered, then I suggest to close this one.

retokromer:
>If #160 can be considered, then I suggest to close this one.

Excellent!

dericed:
Closing in favor of https://github.com/FFmpeg/FFV1/pull/160.

PR #184, opened 2019-12-10
add citations and references for figures and tables referenced within…

dericed:
… the text

This pr starts to add citations for figures and tables. I suggest that this treatment be applied to all figures and tables so that they are properly contextualized and able to be referenced.


dericed:
rebased after https://github.com/FFmpeg/FFV1/pull/185 was merged to resolve conflict. Please re-review.

PR #187, opened 2020-01-24
Definition fixes

michaelni:


ffv1.md
```diff
@@ -71,7 +71,7 @@ Note: the operators and the order of precedence are the same as used in the C pr
 
 `a | b`       means bit-wise "or" of a and b.
 
-`a >> b`      means arithmetic right shift of two’s complement integer representation of a by b binary digits.
+`a >> b`      means arithmetic right shift of two’s complement integer representation of a by b binary digits. This is equivalent to, b times dividing a by 2 with rounding toward negative infinity.
```
dericed: Rewording suggestion:
`This is equivalent to the largest integer less than or equal to a divided by 2 for b times.`

ffv1.md
```diff
@@ -53,7 +53,7 @@ The FFV1 bitstream is described in this document using pseudo-code. Note that th
 
 ### Arithmetic Operators
 
-Note: the operators and the order of precedence are the same as used in the C programming language [@!ISO.9899.2018].
+Note: the operators and the order of precedence are the same as used in the C programming language [@!ISO.9899.2018]. With the exception of `>>` (removial of implementation defined behavior) and `^` (power instead of XOR)
```
dericed: Suggested rewording.
> Note: the operators and the order of precedence are the same as used in the C programming language [@!ISO.9899.2018], with the exception of `a >> b` and `a ^ b` operators which are re-defined within this section.

ffv1.md
michaelni: what definition of divided does this use. its not what C uses as that rounds to 0 and you no longer  change the rounding so i think this is ambigous.

retokromer: I agree that rounding toward negative infinity is important to state.

ffv1.md
michaelni: I think the wording should still contain a summary of the differences. This makes the text easier to understand. And not just correct

dericed: I'd be fine with adding the parenthetical hints back in. My initial motivation to edit was to merge the two sentences and correct a typo ('removial').

michaelni: would that work: ?
"With the exception of `>>` (removal of implementation defined behavior) and `^` (power instead of XOR) operators which are re-defined within this section."

dericed: +1

ffv1.md
```diff
@@ -133,8 +133,11 @@ median(a,b,c)         the numerical middle value in a data set of a, b, and c, i
 
 a~b~                  the b-th value of a sequence of a
 
-a~b,c.                the 'b,c'-th value of a sequence of a
+a~b,c~                the c-th value of a sequence of a~b~
```
mcr: Maybe some additional text, an example would help here.

michaelni: updated the pull req, removing the whole sequence stuff (that is all uses of ~)

PR #188, opened 2020-01-24
Eliminate XOR, i think these where AND a long time ago

michaelni:
Signed-off-by: Michael Niedermayer <michael@niedermayer.cc>


dericed:
This may be my error from https://github.com/FFmpeg/FFV1/commit/d036c0f31d6dc1e6fb6ce0278790fc07ef927e9a when I initially worked to generate an ascii art representation of the latex equation. Not sure why, but apparently I interpreted the wedge as an XOR rather than a logical AND. Thx for finding this.

LGTM

PR #192, opened 2020-03-31
Clarification about Parameters() intra field values

JeromeMartinez:
There may be some misunderstanding about `intra` field in `Parameters`, with some mix with `keyframe` field in `Frame` as seen on CELLAR.
The reference decoder reads this value but ignores it (no impact on the decoding).

This PR clarifies the link between `intra` field in `Parameters` and `keyframe` field in each `Frame`, and make explicit that `intra` value of 1 has no impact on the decoding process (it is only an hint).

### DIFF
--- a/ffv1.md
+++ b/ffv1.md
@@ -1012,14 +1012,14 @@ AART:       ( pred + initial_state_delta[ i ][ j ][ k ] ) & 255
 
 ### intra
 
-`intra` indicates the relationship between the instances of `Frame`.
+`intra` indicates the constraint on `keyframe` in each instance of `Frame`.
 
 Inferred to be 0 if not present.
 
 |value  | relationship                                                     |
 |-------|:-----------------------------------------------------------------|
-|0      | Frames are independent or dependent (keyframes and non keyframes)|
-|1      | Frames are independent (keyframes only)                          |
+|0      | `keyframe` can be 0 or 1 (non keyframes or keyframes)            |
+|1      | `keyframe` MUST be 1 (keyframes only)                            |
 |Other  | reserved for future use                                          |
 
 ## Configuration Record


ffv1.md
```diff
@@ -1018,10 +1018,12 @@ Inferred to be 0 if not present.
 
 |value  | relationship                                                     |
 |-------|:-----------------------------------------------------------------|
-|0      | Frames are independent or dependent (keyframes and non keyframes)|
-|1      | Frames are independent (keyframes only)                          |
+|0      | no constraint expected on `keyframe`                             |
+|1      | `keyframe` is expected to be always 1 (keyframes only)\*         |
 |Other  | reserved for future use                                          |
 
+\* This is only an hint. Decoders MUST NOT rely on this value for guessing the value of `keyframe` in each `Frame`.
```
retokromer: a hint

michaelni:
This change is not ok.
A decoder can use the flag to know that it can decode multiple frames at the same time. To know that it can seek to any frame and decode it. To know it can skip any frame without consequence to the following frames.

JeromeMartinez:
The reference decoder does not check `intra` value (it reads it, but does not use it for e.g. ignoring the `keyframe` value or bypassing it, `keyframe` value is read and prioritized).
For instance, `intra` of 1 and `keyframe` of 0 is currently handled as non key frame (as if `intra` is 0).
So I mapped the reference decoder behavior.

In my mind it means that the `intra` field is an hint only. Else `keyframe` field would just be removed from the bitstream as it is useless (it could be the case in V4, so all slices would be independent, BTW! But it is another story...).

Do you prefer something like if `intra` is 1, `keyframe` of all frames MUST be 1?

michaelni:
> Do you prefer something like if `intra` is 1, `keyframe` of all frames MUST be 1?

yes

JeromeMartinez:
> > Do you prefer something like if intra is 1, keyframe of all frames MUST be 1?

> yes

PR modified accordingly.

JeromeMartinez:
bump ;-)

PR #194, opened 2020-03-31
Fix some formatting issues

JeromeMartinez:
As underscores have a specific meaning in Markdown, not escaping them has some ugly results when converted to HTML, e.g.

ffv1

I tried to catch all fields with underscores, with this rule in mind:
- paragraph titles and table column names have the underscore escaped, not considered as "code", even if only 1 underscore (currently not caught as italic, but I prefer to take no risk for the future in case we change of Markdown processor and this make all titles coherent with always the underscore escaped).
- when used as bitstream fields directly (alone or in a formula), they are always within 2 backquotes (which disables the italic meaning of the undesrcore).

Let me know if you prefer all of the fields within 2 backquotes i.e. also in paragraph titles and table column names.

### DIFF
--- a/ffv1.md
+++ b/ffv1.md
@@ -332,7 +332,7 @@ AART:g=Y-(Cb+Cr)>>2
 AART:r=Cr+g
 AART:b=Cb+g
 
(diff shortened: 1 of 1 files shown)


dericed:
Can you confirm that the cross-sectional links work after this. For instance in section 3.8.1.6 there should be a link to the coder_type section (4.1.3). Does it still work in the txt and html outputs?

JeromeMartinez:
bump ;-)

PR #200, opened 2020-05-03
Adjust appendices

dericed:
This pull request builds upon https://github.com/FFmpeg/FFV1/pull/199 and responds to the following comment at https://mailarchive.ietf.org/arch/msg/cellar/K1girgkVr4AKNDzPcUPyRjla7Rc/:

> Section 9:
> * Appendices go after the references.

PR #202, opened 2020-05-09
More ad review updates take 2

dericed:
This PR is intended to supersede both https://github.com/FFmpeg/FFV1/pull/199 and https://github.com/FFmpeg/FFV1/pull/200. It makes the commits about appendices handling easier to follow.

I agreed with @michaelni's [comment](https://github.com/FFmpeg/FFV1/pull/199#issuecomment-624910280) and used the term "FFV1 bitstream" rather than "a content", though the comment there about the colorspace restriction may still need work. The comment there about line-length is the same before and after the PR (It's about some pseudo-code table being too long), so I can adjust that in a subsequent ping.

Pinging @JeromeMartinez to review as well since this edits his work.

### DIFF
--- a/ffv1.md
+++ b/ffv1.md
@@ -922,7 +922,7 @@ If `state_transition_delta` is not present in the FFV1 bitstream, all Range code
 
 Restrictions:
 
-If `colorspace_type` is 1, then `chroma_planes` MUST be 1, `log2_h_chroma_subsample` MUST be 0, and `log2_v_chroma_subsample` MUST be 0.
+Decoders SHOULD reject FFV1 bitstreams with `colorspace_type` == 1 && (`chroma_planes` != 1 || `log2_h_chroma_subsample` != 0 || `log2_v_chroma_subsample` != 0).
 
 ### chroma\_planes
 
--- a/ffv1.md
+++ b/ffv1.md
@@ -1172,8 +1172,6 @@ Encoders SHOULD NOT fill these bits.
 
 Decoders SHOULD ignore these bits.
 
-Note in case these bits are used in a later revision of this specification: any revision of this specification SHOULD care about avoiding to add 40 bits of content after `SliceContent` for `version` 0 and 1 of the bitstream. Background: Due to some non-conforming encoders, some bitstreams were found with 40 extra bits corresponding to `error_status` and `slice_crc_parity`. As a result, a decoder conforming to the revised specification could not distinguish between a revised bitstream and a buggy bitstream.
-
 ## Slice Header
 
 A `Slice Header` provides information about the decoding configuration of the `Slice`, such as its spatial position, size, and aspect ratio. The pseudo-code below describes the contents of the `Slice Header`.
--- a/ffv1.md
+++ b/ffv1.md
@@ -857,7 +857,7 @@ Decoders SHOULD reject FFV1 bitstreams with version >= 3 && ConfigurationRecordI
 |4       |  FFV1 version 4         |{V4}
 |Other   |  reserved for future use|
 
-\* Version 2 was never enabled in the encoder thus version 2 files SHOULD NOT exist, and this document does not describe them to keep the text simpler.
+\* Version 2 was never enabled in any encoder known to the authors of this document. Decoders SHOULD reject FFV1 bitstreams with `version` == 2.
 
 ### micro\_version
 
--- a/ffv1.md
+++ b/ffv1.md
@@ -1575,12 +1575,6 @@ The IANA is requested to register the following values:
 
    - Media type registration as described in (#media-type-definition).
 
-# Appendix A: Multi-theaded decoder implementation suggestions
-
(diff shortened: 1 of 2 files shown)


dericed:
Nudging on this to complete the response to the review. Or in regards to the comments at https://github.com/FFmpeg/FFV1/pull/199#issuecomment-624910280 I could move out the "More explicit wording ..." commits for separate consideration.

JeromeMartinez:
Would it be possible to merge and/or comment about the PR before the CELLAR meeting next Tuesday?

michaelni:
> -\* Version 2 was never enabled in the encoder thus version 2 files SHOULD NOT exist, and this document does not describe them to keep the text simpler.
> +\* Version 2 was never enabled in any encoder known to the authors of this document. Decoders SHOULD reject FFV1 bitstreams with `version` == 2.

"Decoders SHOULD reject FFV1 bitstreams with `version` == 2"  <-- This makes no sense. This document does not describe the version 2 bitstream nor how a encoder would create a version 2 bitstream. Why would it describe what a decoder should do with it, and then recommand not to decode it. 
Thats as if jpeg2000 spec said, we dont describe classic jpeg but if you encounter one you should reject it. 
I fail to understand why that kind of recommandition should be in ... me.is_confused() == true

> -If `colorspace_type` is 1, then `chroma_planes` MUST be 1, `log2_h_chroma_subsample` MUST be 0, and `log2_v_chroma_subsample` MUST be 0.
> +Decoders SHOULD reject FFV1 bitstreams with `colorspace_type` == 1 && (`chroma_planes` != 1 || `log2_h_chroma_subsample` != 0 || `log2_v_chroma_subsample` != 0).

This confuses me.
Either the case described here is allowed, in which case it is not described what RGB without chroma would mean. Or it is not allowed then its not allowed for a encoder to generate that and only could be encountered if a stream was damaged before a decoder. And the suggestion of rejecting it does not really help handling damaged streams

michaelni:
Applied most commits of this pull request

JeromeMartinez:
> This document does not describe the version 2 bitstream nor how a encoder would create a version 2 bitstream.

Would it be fine with:
"Version 2 was experimental and this document does not describe it."

this way we still have an informative sentence and at the same time with don't have the "SHOULD NOT" which is weird here as it is not descriptive of the spec.

> Or it is not allowed then its not allowed for a encoder to generate that and only could be encountered if a stream was damaged before a decoder.

It is not allowed but we need to find a way to describe from a decoder point of view and not from an encoder point of view. IMO a decoder should reject such bitstream as it is not decodable / it does not make sense, and the handling of damaged streams is not part of the document.

Suggestion:
"Decoders conforming to this version of this specification SHALL ignore bitstreams with colorspace_type == 1 && (chroma_planes != 1 || log2_h_chroma_subsample != 0 || log2_v_chroma_subsample != 0)."

Or any suggestion?

retokromer:
>Would it be fine with:
"Version 2 was experimental and this document does not describe it."

I would add an “only”:
"Version 2 was experimental only and this document does not describe it."

>Suggestion:
"Decoders conforming to this version of this specification SHALL ignore bitstreams with colorspace_type == 1 && (chroma_planes != 1 || log2_h_chroma_subsample != 0 || log2_v_chroma_subsample != 0)."

This sounds very weird to me, but I don’t have a suggestion.

michaelni:
> > This document does not describe the version 2 bitstream nor how a encoder would create a version 2 bitstream.
> 
> Would it be fine with:
> "Version 2 was experimental and this document does not describe it."
> 
> this way we still have an informative sentence and at the same time with don't have the "SHOULD NOT" which is weird here as it is not descriptive of the spec.

ok


> 
> > Or it is not allowed then its not allowed for a encoder to generate that and only could be encountered if a stream was damaged before a decoder.
> 
> It is not allowed but we need to find a way to describe from a decoder point of view and not from an encoder point of view. IMO a decoder should reject such bitstream as it is not decodable / it does not make sense, and the handling of damaged streams is not part of the document.
> 
> Suggestion:
> "Decoders conforming to this version of this specification SHALL ignore bitstreams with colorspace_type == 1 && (chroma_planes != 1 || log2_h_chroma_subsample != 0 || log2_v_chroma_subsample != 0)."
> 
> Or any suggestion?

I would make this more generic, something like
This specification describes the valid bitstream and how to decode such valid bitstream. Bitstreams not conforming to this specification or how they are handled is outside this specification. A decoder could reject every invalid bitstream or attempt to perform error concealment or re-download or use a redundant copy of the invalid part or any other action it deems appropriate.

JeromeMartinez:
> I would make this more generic, something like

I get the idea but I IMO I don't see how to put it for the exact line about `colorspace_type == 1` limitations.
I have created a new PR https://github.com/FFmpeg/FFV1/pull/203 with a tentative of wording about `colorspace_type == 1` limitations matching the idea described + another patch with your wording in introduction, @michaelni please review the new PR, @dericed I think we can close this PR.

dericed:
thx closing

PR #205, opened 2020-05-26
keep equation within line length constraints

dericed:
After @JeromeMartinez 's commit at https://github.com/FFmpeg/FFV1/commit/9228942500e6bc92943757b3a975255de7a72465 there is an idnit error as the line length exceeds 72 characters in the rendered txt output. This is an attempt to get the output back to <=72.

### DIFF
--- a/ffv1.md
+++ b/ffv1.md
@@ -374,8 +374,8 @@ SVGI:!---
 SVGI:![svg](samplediff.svg "coding of the sample difference")
 SVGI:!---
 SVGC:samplediff.svg=$$coder\\\_input=[(sample\\\_difference+2^{bits-1})\\&(2^{bits}-1)]-2^{bits-1}$$
-AART:coder_input =
-AART:    [(sample_difference + 2 ^ (bits-1)) & (2 ^ bits - 1)] - 2 ^ (bits - 1)
+AART:coder_input = [(sample_difference + 2 ^ (bits - 1)) &
+AART:              (2 ^ bits - 1)] - 2 ^ (bits - 1)
 
 ### Range Coding Mode
 


ffv1.md
```diff
@@ -375,7 +375,7 @@ SVGI:![svg](samplediff.svg "coding of the sample difference")
 SVGI:!---
 SVGC:samplediff.svg=$$coder\\\_input=[(sample\\\_difference+2^{bits-1})\\&(2^{bits}-1)]-2^{bits-1}$$
 AART:coder_input =
-AART:    [(sample_difference + 2 ^ (bits-1)) & (2 ^ bits - 1)] - 2 ^ (bits - 1)
+AART:[(sample_difference + 2^(bits-1)) & (2^bits - 1)] - 2^(bits - 1)
```
retokromer, line 378: I suggest to unify the two `(bits-1)` without spaces around the minus sign. In my opinion, it makes the last one much more readable, especially considering the different one in the middle.

retokromer:
>the remaining `bit-1` is a typo

Ah, okay, thank you for clarifying.

>we try another method for being in the 72 character limit

On two lines?

dericed:
updated. The equation now renders as:

```
   coder_input = [(sample_difference + 2 ^ (bits - 1)) &
                 (2 ^ bits - 1)] - 2 ^ (bits - 1)

                                  Figure 7
```

michaelni:
would a M = 2 ^ bits - 1 make it more readable and fit ?

dericed:
I don't mind the line break, we have it in a few other places such as https://datatracker.ietf.org/doc/html/draft-ietf-cellar-ffv1-13#section-3.4 and https://datatracker.ietf.org/doc/html/draft-ietf-cellar-ffv1-13#section-3.8.1.1. 

Alternatively, I could make it like:
```
coder_input = [(sample_difference +
               2 ^ (bits - 1)) &
               (2 ^ bits - 1)] -
               2 ^ (bits - 1)
```

dericed:
nudging here. I think I still prefer the update as it is (as shown in https://github.com/FFmpeg/FFV1/pull/205#issuecomment-634068010).

I tried a draft of @michaelni's proposal as it was like
```
M = 2 ^ bits - 1
coder_input = [(sample_difference + 2 ^ (bits - 1)) & M] - 2 ^ (bits - 1)
```

Taken further though, it could be:
```
M = 2 ^ bits - 1
N = 2 ^ (bits - 1)
coder_input = [(sample_difference + N) & M] - N
```

No strong opinion, rather than moving along this PR, but I prefer the PR as it is.

retokromer:
>nudging here. I think I still prefer the update as it is (as shown in https://github.com/FFmpeg/FFV1/pull/205#issuecomment-634068010).

I also prefer this one.

The only change I would personally made is to put the `&` at the beginning of the new line rather than at the end of the “old” one. To me it’s more readable (possibly an atavism of my maths past).

PR #210, opened 2020-05-27
QuantizationTableSet place

JeromeMartinez:
More AD review updates, it is preferred to define `QuantizationTableSet` before using it, so I move the corresponding section (no change in the section itself).

I didn't inverted `QuantizationTableSet` and `QuantizationTable` because the remark was focused on `QuantizationTableSet` being a lot later than when it is used.

### DIFF
--- a/ffv1.md
+++ b/ffv1.md
@@ -784,6 +784,64 @@ Default values at the decoder initialization phase:
 
 `ConfigurationRecordIsPresent` is set to 0.
 
+## Quantization Table Set
+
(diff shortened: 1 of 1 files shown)


ffv1.md
```diff
@@ -786,6 +786,62 @@ Default values at the decoder initialization phase:
...
+`MAX_CONTEXT_INPUTS` is 5.
+
+```c
+pseudo-code                                                   | type
+--------------------------------------------------------------|-----
+QuantizationTable(i, j, scale) {                              |
+    v = 0                                                     |
+    for (k = 0; k < 128;) {                                   |
+        len - 1                                               | ur
+        for (a = 0; a < len; a++) {                           |
```
retokromer, line 821: Is `a` necessary?

JeromeMartinez: @retokromer how do you imagine to say in C pseudo-code that the loop has `len` iterations? I don't see how without this `a`.

retokromer: I know that it’s an iteration, but I would say `n` rather than `a` which has usually other meanings.

retokromer: … or possibly even better `l`.

JeromeMartinez:
I rebased.
@michaelni note that the PR is no more without change in the section: the first commit is with no change in the section itself, the 2nd commit is a change.

PR #211, opened 2020-05-27
Provide links for describing how put_symbol is used

JeromeMartinez:
(I hope!) Last AD review update, removing "followed by some comments" and providing a link for `put_rac`.


JeromeMartinez:
nudge about the 4 remaining PRs to merge, if we can have a version ready to go before the next IETF meeting...

PR #214, opened 2020-06-23
bump to ffv1 15 and ffv1 v4 12

dericed:
I suggest addressing https://github.com/FFmpeg/FFV1/pull/212 first as master is broken.


dericed:
Thx @michaelni! Can you mark a release in the repo as well. I'll update the IETF data tracker.

PR #215, opened 2020-06-25
Equations review

dericed:


dericed:
Noting that the cellar AD okayed this at https://mailarchive.ietf.org/arch/msg/cellar/lwJM5d6CH7kXGQaURN8E1UwCSSM/

PR #222, opened 2020-07-29
add buildcheck system for checking dependencies

dericed:
This PR follows conversations with https://github.com/FFmpeg/FFV1/pull/219#issuecomment-658229944 and expands on work by @robUx4 in the https://github.com/cellar-wg/ebml-specification repository. Both repositories rely on rfc2xml and mmark, and this pr adds in installs for xmlstarlet, pdfcrop, pdf2svg. Comments welcome.


michaelni:
I would avoid a dependancy on specific shell like bash.
but the much bigger problem is tools.mak, this is just a huge security hole.
As I said previously i dont think the script should install things but just tell the admin how to do it.
If you really want a installer script, it needs to inform the admin what it will do, let him confirm. Maintain a list of file hashes of things, so that each tar.gz thats downloaded would have its sha256 checked against a list so a trojanized version would be rejected. Also pip and all similar tools AFAIK have the issue that if you update the OS you might have a newer package at OS level than user level and break all kinds of things of the OS for that user in mysterious ways.

IMHO the best is not to depend on tools not widely available unless mandated by IETF.
widely available tools are easy to install by using the OS/Platform package management or equivalent.
When a version of a tool is installed that is not widely available, the admin must be aware of it because that can have side effects on other thing than the ffv1 build. The admin may choose to use a seperate user, a VM, a different package repository than the official one to install a needed package. The partition the git repo is on may also be mounted noexec making any automated install fail.
A script that download code and executes it is a big red security flag. Its something that may make sense as a last resort in rare cases but this is a repository for writing a document, i dont understand why we would need or want this.
Document what tools are needed and why and help the admin by telling him how he can securely and saftely install these versions.  I think would be a very good idea. the tools script/Makefile can just tell the admin what to do instead of doing it. But automated install is a risky thing. And yes i know 90% of people probably dont care that much but we should take security ....

robUx4:
The tools.mak is derived from the one we use in VLC for compiler tools (pretty much everything except the compiler comes in). It does have a sha512 system to verify the downloaded version matches the one expected. So this could be added.

One reason to be able to download some tools is that they can be used in Github Actions to process each push/PR without modifying github actions itself. For example you may need to update a tool version. But it's also [possible to do it directly in Actions](https://github.com/cellar-wg/matroska-specification/blob/master/.github/workflows/generate-rfcs.yaml#L63), just a bit more work. It's also a bit harder if the tools is not available via Actions (in this case no Ubuntu or package).

dericed:
I updated the PR to avoid the security concerns, now there's simply a script to check the presence and versions of the installed dependencies.

dericed:
closing. This PR was partly superseded by https://github.com/FFmpeg/FFV1/pull/228.

PR #232, opened 2020-09-24
simplify artwork process

dericed:
this takes advantage of some recent methods in mmark to handle figures

The output is the same except some artwork types which were unintentionally added via svg2src are now not added. Also the process this branch uses adds a new line after ascii-art which was previously stored after `AART:`.


michaelni:
Iam not sure i understand what this improves ?
The AART/SVG prefixes used previously seemed somewhat self explanatory

dericed:
The PR moves some aspects of handling ascii-art to methods native to mmark rather than our own custom work, but true that the current method is self-explanatory. Will close.

PR #235, opened 2020-09-30
mention pandoc as dependency

dericed:
related to https://github.com/FFmpeg/FFV1/issues/234#issuecomment-701504192


michaelni:
ubuntu 18.04LTS has only pandoc 1.19.2.4
What is required thats not suppored by that ?

dericed:
I tested and 1.19.2.1 works (svg outputs match) so I reduced the requirement to that.

PR #237, opened 2020-10-01
Makefile: add xml2rfc.version-ok xmlstarlet.version-ok pdfcrop.version-ok tests

michaelni:


dericed (approved):
tested and this works for me. Thx so much!

dericed:
oh, these .version-ok files should be added to .gitignore

retokromer (approved):

PR #238, opened 2020-10-02
Ballot Barry part 1

JeromeMartinez:
A couple of patches partially fixing [Barry Leiba's Discuss](https://mailarchive.ietf.org/arch/msg/cellar/rvI2_U8IydKtrOo1270wVOddyug/).

PR #239, opened 2020-10-02
Remove quotes from defined words 

JeromeMartinez:
From [Barry Leiba's Discuss](https://mailarchive.ietf.org/arch/msg/cellar/rvI2_U8IydKtrOo1270wVOddyug/):

> General: It’s jarring to see the defined terms, such as “Sample” and “Pixel” in
quotes every time they’re used.  More common practice is to quote them when
they’re first defined, and then to just use the capitalization to show that
they’re defined terms.  I find that the quotes affect the readability and make
it somewhat more of a challenge to read.

So I removed quotes from defined words.
As `Frame` is not only a defined word but also a part of the bitstream, it is not just a find and replace, so maybe some double check to do.

I also added the missing definition of `Line`.

### DIFF
--- a/ffv1.md
+++ b/ffv1.md
@@ -25,27 +25,27 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT", "SHOULD", "S
 
 ## Definitions
 
-`Container`: Format that encapsulates `Frames` (see (#frame)) and (when required) a `Configuration Record` into a bitstream.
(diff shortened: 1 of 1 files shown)


retokromer (approved):

michaelni:
Does not apply cleanly to master/HEAD anymore "This branch has conflicts that must be resolved"

JeromeMartinez:
PR udpated for including changes from https://github.com/FFmpeg/FFV1/pull/238, no more conflicts.

Issue #230, opened 2020-09-22
section 3.8.1.2 needs to define get_rac


mcr:
As per Barry Lieba review, the section 3.8.1.2 says:

     "get_rac" returns a boolean, computed from the bytestream as
     described in Section 3.8.1.1. 

but it is not defined, and likely relates to the decode process.

mcr:
https://github.com/dwbuiten/go-ffv1/blob/master/ffv1/rangecoder/range.go  
also change reference from 3.8.1.1 to cite figure 9.

mcr:
@dericed

dwbuiten:
Worth noting here that when I implemented go-ffv1, it was with an older revision of the spec, and it seems the figure numbers have since changed... I'll update the comments.

(Also worth noting that go-ffv1 is not an official reference decoder.)

dericed:
Hi @dwbuiten, you could contextualize your comments to the version of the draft such as "draft-ietf-cellar-ffv1-17", although as the document is in "IESG Evaluation" status the RFC number may be soon.

dwbuiten:
Yes, I'll do that when I update them.

dericed:
@dwbuiten, @michaelni, are you able to assist with the get_rac piece of Barry's review. The quote from the [review](https://mailarchive.ietf.org/arch/msg/cellar/rvI2_U8IydKtrOo1270wVOddyug/) is 

```
— Section 3.8.1.2 —

   "get_rac" returns a boolean, computed from the bytestream as
   described in Section 3.8.1.1.

I see nothing in Section 3.8.1.1 that describes get_rac.
```

I understand that get_rac refers to [figure 9](https://www.ietf.org/id/draft-ietf-cellar-ffv1-17.html#figure-9); however a narrative could help. For instance does `rac` mean `RAnge Coded binary value`? If figure 9 is read as `the i-th range coded binary value equals`, then what is the `R_i` of the formula mean?

michaelni:
> I understand that get_rac refers to [figure 9](https://www.ietf.org/id/draft-ietf-cellar-ffv1-17.html#figure-9); however a narrative could help. For instance does `rac` mean `RAnge Coded binary value`?

IIRC get_rac() was named for consistancy with get_cabac(), it returns a single bit coded using the range coder. Not sure if the 3 letters can be matched to grammatically 100% fitting words

JeromeMartinez:
[...]

PR #241, opened 2020-10-03
Ballot Barry part 2

JeromeMartinez:
A couple of patches partially fixing [Barry Leiba's Discuss](https://mailarchive.ietf.org/arch/msg/cellar/rvI2_U8IydKtrOo1270wVOddyug/).

### DIFF
--- a/ffv1.md
+++ b/ffv1.md
@@ -734,7 +734,7 @@ Table: Examples of decoded Signed Golomb Rice Codes. {#tableGolombExamples}
 
 #### Run Mode
 
-Run mode is entered when the context is 0 and left as soon as a non-0 difference is found. The level is identical to the predicted one. The run and the first different level are coded.
+Run mode is entered when the context is 0 and left as soon as a non-0 difference is found. The sample difference is identical to the predicted one. The run and the first different sample difference are coded as defined in (#golomb-rice-sample-difference-coding).
 
 ##### Run Length Coding
 
@@ -833,7 +833,7 @@ get_vlc_symbol(state) {
 }
 ```
 
-##### Level Coding
+##### Golomb Rice Sample Difference Coding
 
 Level coding is identical to the normal difference coding with the exception that the 0 value is removed as it cannot occur:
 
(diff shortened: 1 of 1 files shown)


dwbuiten:
Is it useful to mention `%` means C-style modulo, and no "math-style" modulo?

JeromeMartinez:
> Is it useful to mention `%` means C-style modulo, and no "math-style" modulo?

Checking if we use  `%` on negative numbers, I see that it is actually not used (or did I miss something?), so a way to avoid to have to be more precise is to just remove it from the Order of Operation Precedence :).
PR updated with removal of `%` in Order of Operation Precedence.

ffv1.md
```diff
@@ -672,6 +672,8 @@ ESC MUST NOT be used if the value can be coded as non ESC.
 
 ##### Examples
 
+Below are pratical examples of how Signed Golomb Rice Codes are decoded based on the serie of bits extracted from the bitstream as described by the method above:
```
dericed, line 675: I suggest (fixes 2 spelling errors and adds a table cross-reference
> [@tableGolombExamples] shows practical examples of how Signed Golomb Rice Codes are decoded based on the series of bits extracted from the bitstream as described by the method above:

And add this line after the table:
> Table: Examples of decoded Signed Golomb Rice Codes. {#tableGolombExamples}

JeromeMartinez: Commit updated, @michaelni please review/merge.

retokromer (approved):

dericed (approved):
I re-reviewed and this looks good. Is this PR is the last part of responding to [Barry Leiba's review](https://mailarchive.ietf.org/arch/msg/cellar/Bj-ei6Hyv1ep2k-u8STOMrGx-d0/)?

dericed (approved):

JeromeMartinez:
The only missing part is the get_rac stuff.

dericed:
@JeromeMartinez I'll move the get_rac discussion to [that issue](https://github.com/FFmpeg/FFV1/issues/230#issuecomment-704319546) as I think this PR is ready for merge. hoping @michaelni agrees :)

PR #245, opened 2020-10-06
cross-references for get_rac

dericed:
This adds 2 cross-references to a line about `get_rac`, one referencing a depicting of it as a mathematical formula and one referencing the same in pseudo-code. Ping to @michaelni and @dwbuiten to verify that these cross-references and the captions are fitting.

PR #246, opened 2020-10-06
bump to ffv1 18 and ffv1 v4 15

dericed:
31 commits since last release, so :)


dericed:
I'd suggest merging https://github.com/FFmpeg/FFV1/pull/245 first since it might close Barry's review, but not blocking.

PR #249, opened 2020-10-07
[WIP] Ballot Roman, security part

JeromeMartinez:
From [Roman Danyliw comment](https://mailarchive.ietf.org/arch/msg/cellar/WE0h53hncmrbul1ZoSiSyWEfAeM/):

> On the security considerations:
** Section 6.  Per the reference to [RFC4732], which selection is relevant
here? Is it Section 2.1.1?  If so, the risks due to end-point compromise are
much broader than DoS.
** Section 6.  The assertions about the security properties of [REFIMPL] don’t
make sense to me in this document.  While it is extremely helpful that there is
a high-quality reference implementation, it’s relevance to this spec isn’t
clear.  This code isn’t normative.  Recommend removal all text after the
paragraph “None of the content carried in FFV1 is intended to be executable”.

I have no strong opinion on that, but I was liking the REFIMPL related lines to remove.
What should be done about the RFC4732 remark?

Flagged as "WIP" as I wish input from other writers rather than having this PR merged.


dericed:
This text was added back in https://github.com/FFmpeg/FFV1/commit/41cc0f231d272597150e4df08c75f881804fe0d8. I don't object to it being removed, but I think it would be helpful to communicate that a reference implementation exists. Also note that in this PR you remove a section that contains unique references but the references are then unreferenced.

There are a few mentions of `all known implementations` within the document, perhaps we could have an appendix that clarifies what those are and then in that appendix can contextualize the FFmpeg implementation.

retokromer:
>but I think it would be helpful to communicate that a reference implementation exists

I second this.

JeromeMartinez:
> but I think it would be helpful to communicate that a reference implementation exists.

Same. So should we just answer that we disagree?

JeromeMartinez:
https://github.com/FFmpeg/FFV1/pull/250 is preferred, closing.

PR #255, opened 2020-12-04
move range coder variables to a definition list

dericed:


dericed:
as discussed in https://mailarchive.ietf.org/arch/msg/cellar/sENj_NU60EcmGAsJLnTiYXOONFw/.  I'm using here DefinitionLists defined at https://github.com/gomarkdown/markdown/blob/master/README.md#extensions, which in the output use the <dl> or definition list tag per https://tools.ietf.org/html/rfc7991#section-2.20.

retokromer (approved):

JeromeMartinez (approved):

PR #263, opened 2021-05-25
bump to ffv1 v4 18

dericed:
As requested in the cellar working group meeting. The version of ffv1-v4 needs a bump as the current document is about to expire.


dericed:
Please merge and mark a release. I'll update the IETF data tracker.

dericed:
Ah, hold on this I see that ffv1-v4 version 17 was never uploaded to the datatracker. I'll close this and update the datatracker which will renew the document.

PR #264, opened 2021-06-11
Auth48 changes

dericed:
This is a part one of work to align the xml output of this repository to bring in the corrections of the auth48 process for what will be RFC9043.

PR #265, opened 2021-07-03
More figures and tables titles

JeromeMartinez:


dericed (approved):
Looks good to me. Some redundancy but I agree that having captions for tables and figures is a good goal and in a few places, they more a greater level of clarity. I recommend merging this one first and I'll rebase and clean up https://github.com/FFmpeg/FFV1/pull/264 when the AUTH48 is complete.

dericed:
https://www.rfc-editor.org/auth48/rfc9043 very close :)

ffv1.md
```diff
@@ -288,6 +288,7 @@ SVGI:![svg](quantizationtablesets.svg "quantization table sets")
 SVGI:!---
 SVGC:quantizationtablesets.svg=$$Q_{j}[k]=quant\\_tables[i][j][k\\&255]$$
 AART:Q_(j)[k] = quant_tables[i][j][k&255]
+Figure: Description of the coding of the Quantized Sample Difference. {#figureQuantizationTableSets}
```
michaelni, line 291: This can be read as Quantized (Sample Difference) and (Quantized Sample) Difference, this should be worded so as to be free of this double interpretation

ffv1.md
```diff
@@ -304,6 +305,7 @@ AART:          Q_(1)[tl - t] +
 AART:          Q_(2)[t - tr] +
 AART:          Q_(3)[L - l]  +
 AART:          Q_(4)[T - t]
+Figure: Description of the coding of the Context. {#figureContext}
```
michaelni, line 308: Coding generally to me means storing in the bitstream, we do not really store the context. Rather the Context for each sample is a consequence of the previously stored samples.

ffv1.md
```diff
@@ -1225,13 +1239,15 @@ SVGI:![svg](initialstatedelta1.svg "initial state delta 1")
 SVGI:!---
 SVGC:initialstatedelta1.svg=pred = j ? initial\_states[ i ][j - 1][ k ] : 128
 AART:pred = j ? initial_states[ i ][j - 1][ k ] : 128
+Figure: Predicator value for the coding of `initial_state_delta[ i ][ j ][ k ]`. {#figureInitialStateDeltaPred}
```
michaelni, line 1242: What is a "Predicator"?

ffv1.md
JeromeMartinez: "Quantized Sample Difference" is the wording used in the previous paragraph (actually already 6 times in the spec), should it be changed everywhere? It is a bit much as we are in the final part of the RFC, I am reluctant to change everywhere or use a wording not same as elsewhere.

ffv1.md
JeromeMartinez: `coding` changed to `computing`.

ffv1.md
JeromeMartinez: Typo `Predictor` fixed.

ffv1.md
michaelni: does each occurance of "Quantized Sample Difference" refer to the same thing ?

JeromeMartinez: > does each occurance of "Quantized Sample Difference" refer to the same thing ?

All but one are at the same place (in ["Quantization Table Sets" part](https://www.rfc-editor.org/authors/rfc9043.html#name-quantization-table-sets) or just after, in "Context"), the other one is in ["quant_tables" part](https://www.rfc-editor.org/authors/rfc9043.html#name-quant_tables) which refers to the same topic, as I understand.

dericed:
The updates look good to me.

dericed:
I see that Jean updated the xml. I'll offer that once this PR is merged, I'll reply to Jean with directions on what edits to make to rfc9043.xml to correlate to this PR. Is any further discussion here needed though?

ffv1.md
michaelni: The quantization table converts sample differences in quantized (sample differences). But the storage of the quantization table has no samples, it stores really the quantization thresholds or levels or their differences. So iam a bit confused by the wording and what "Quantized Sample Difference" refers to here

dericed: From my reading of Jerome's PR, I presumed that "Quantized Sample Difference" referred back to the expression of the same phrase earlier in the section.
> Quantization Tables are used on Sample Differences (see (#coding-of-the-sample-difference)), so Quantized Sample Differences are stored in the bitstream.

however, as the figure's label demonstrates how to compute the index (least 8 bits of the Quantized Sample Difference).

Rather than:
> Figure: Description of the coding of the Quantized Sample Difference. {#figureQuantizationTableSets}

how about
> Figure: Description of the coding of the index of the Quantized Sample Difference. {#figureQuantizationSampleDifferenceIndex}

Here the caption simply refers to the same terminology and assertions of the sentence that introduces the formula.

JeromeMartinez: I modified the patch accordingly.

dericed (approved):
LGTM, thanks Jerome. I'll await @michaelni's review and if merged, I'll prep a change request for the AUTH48 editors.

retokromer (approved):

ffv1.md
michaelni: > From my reading of Jerome's PR, I presumed that "Quantized Sample Difference" referred back to the expression of the same phrase earlier in the section.
> 
> > Quantization Tables are used on Sample Differences (see (#coding-of-the-sample-difference)), so Quantized Sample Differences are stored in the bitstream.

Where are the Quantized Sample Differences stored ?

If i look at the text its not clearly worded
> Relative to any Sample `X`, the Quantized Sample Differences `L-l`, `l-tl`, `tl-t`, ` T-t`, and `t-tr` are used as context:
 `L-l`, `l-tl`, `tl-t`, ` T-t`, and `t-tr`  are Sample Differences, they become Quantized when passing through the quantization table, thats why its called a quantization table.

Maybe iam missing something but if not then the followig sounds more correct and maybe something like this could be used:
Figure: Description of the mapping from sample differences to the corresponding Quantized Sample Differences

dericed: I updated @JeromeMartinez's patch per @michaelni's proposal.

dericed: Acknowledging the @JeromeMartinez is afk but okays change with google hangouts discussion.

JeromeMartinez: Just retrieved my credentials :). Fine for me.

michaelni:
I merged this so we can move forward but maybe some of this text can be worded better, i think there are still 2 occurances of "coded" in there that would be better to be worded differently

dericed:
Any proposal for a different term? The next step is for the 3 authors to okay rfc9043.xml after the figures are added, so it's really the last chance.

michaelni:
I hope i can stiill think clearly after all that terminology lawyering that cant be healthy. but i would say:
Definition of the  state transition table for a `get_rac` readout value of 0/1
I also notice that one_state is connexted to 0 and zero to 1, maybe thats correct but it just caught my eye

PR #270, opened 2022-05-24
Table fix and version bump

dericed:


dericed:
@michaelni, if you can merge this and make another release I can update the tracker and we'll be in sync. I think for the last iteration I updated the tracker and missed the bump here so we fell out of sync.

michaelni:
I can merge this but i just bumped the versions and did a release a few hours ago on github isnt that good enough for adding to that tracker thing ?
I can merge this afterwards
but maybe iam missing something

dericed:
I think I had updated the version number locally but not in github in order to quickly post a new version in a meeting, which is why an extra bump is currently required. With the bump in this PR, the repo builds the same version number as currently in the ietf tracker.

PR #271, opened 2022-12-05
bump version 4 to 20 and 0-3 to 23

dericed:
Per https://www.ietf.org/rfcdiff?url2=draft-ietf-cellar-ffv1-v4-20. This is just a bump as version 19 in the ietf tracker was expired.


dericed:
and reminder to mark release in https://github.com/FFmpeg/FFV1/releases as well

